PR #780, opened 2013-03-23
Verify that the csv data is saved in UTF-8 format before you import this...

blacklight2000:
... data into mongodb.


mpaik:

PR #1470, opened 2013-12-12
Changed Earth radius to reflect the constant used by MongoDB

rps:
MongoDB sets a constant defining the Earth's radius in km as 6378.1. Documentation should reflect this to ensure that users are consistent in their conversion from km to rad.


mpaik:

rps:
The net effect of having documentation inconsistent with the constants used by Mongo is that end users will conduct inaccurate conversions from kilometers to radians. For geospatial applications reliant on Mongo, this results in inaccurate and incomplete proximity search. While I understand that changing the constant used by Mongo breaks existing indexes, I still don't understand why the documentation does not reflect the variable used.

mpaik:
Merged.

PR #1536, opened 2014-01-17
Proper way to get Windows version

jamestyj:
The original command here only returns the architecture (e.g. 64-bit) .
This command will return the version, build number, and architecture.
Tested on Windows Server 2012.


mpaik:

PR #1634, opened 2014-03-06
Update install-mongodb-on-windows.txt

biniama:
I added an example of how the data\db folder can be accessed from the command line passing --dbpath argument.


mpaik:

PR #1766, opened 2014-05-08
Disable balancer while indexes are inconsistent

achille:
A chunk migration can cause an index build, so we should advise to disable the balancer to avoid performance issues.


mpaik:

PR #1830, opened 2014-06-09
Fix echo command for creating config file on Windows

kuc:
Previous command was creating a file with content (literally):
```
logpath="C:\Program Files\MongoDB\log\mongo.log"
```
which was invalid, as reported by `mongod` at startup.
Should be
```
logpath=C:\Program Files\MongoDB\log\mongo.log
```


mpaik:

PR #1870, opened 2014-07-02
Update configure-replica-set-secondary-sync-target.txt

sflint:
This is confusing .... adding in help to make it less confusing


mpaik:

PR #2058, opened 2014-10-31
Clarify when $elemMatch is not necessary

dandv:
Please correct if wrong. Other nuances of $elemMatch vs. array.field would be helpful.

### DIFF
--- a/source/reference/operator/query/elemMatch.txt
+++ b/source/reference/operator/query/elemMatch.txt
@@ -30,8 +30,8 @@ for :query:`$elemMatch`.
 Examples
 --------
 
-Element Match
-~~~~~~~~~~~~~
+Primitive type element match
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Given the following documents in the ``scores`` collection:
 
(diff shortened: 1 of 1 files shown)


dandv:
Any news?

kay-kim:
Hi Dan -- 
Thank you for the pull request as well as the follow up and apologies for the lateness of this reply.  We will get to your pull request sometime next week.  

Kind regards,

Kay  Kim

kay-kim:
Hi Dan -- 
will merge this pull request tomorrow; however, I just wanted to give you a heads up that I may pop the information as a footnote, similar to what we have in the $all page
http://docs.mongodb.org/manual/reference/operator/query/all/#illustrative, as opposed to a separate example.

Otherwise, people querying would have to know a priori that all documents in the collection only contain and will forever only contain at most 1 element in that array field, in which case, the use of the array itself seems unnecessary for the field.

kay-kim:
Hi Dan -- 
I think I misunderstood what your example was illustrating. Is your example to highlight that if you only have one predicate, you don't need to use $elemMatch?

kay-kim:
Merged.  I further edited $elemMatch to clarify that with only a single query predicate in the $elemMatch expression, you don't need to use $elemMatch. I decided not to footnote that if you only have a single element in an array because of the aforementioned caveats.  Thanks again.

Regards,

Kay Kim

PR #2082, opened 2014-12-13
Update db.collection.update.txt

sam-at-github:
reword ambiguous sentence on update parameters. Also add "<>" around to update method parameters in synopsis, inline with (most) other synopsis.


mpaik:

PR #2090, opened 2014-12-21
SERVER-16550 non-text keys in compound text indexes can't be used to sor...

ramonfm:
...t


kay-kim:
Hey Ramon -- will close in favor of https://github.com/mongodb/docs/commit/7883c51560ac9232c0c0dc732d8c760326585486  for this since you just can get sort order from text index compound or not.  Feel free to tweak as you wish.

PR #2096, opened 2015-01-05
DOCS-4522: adds new mongotop error message for 2.8

schmalliso:
Also adds a general note that mongotop throws an error when the primary isn't available and it's connected to   replica set.

LGTM'd by Wisdom.

### DIFF
--- a/source/includes/options-mongotop.yaml
+++ b/source/includes/options-mongotop.yaml
@@ -36,6 +36,17 @@ inherit:
   name: host
   program: _shared
   file: options-shared.yaml
+post: |
+
+
+  .. versionchanged:: 2.8.0
+     If connected to a replica set, {{program}} only reports activity
+     on the :term:`primary`. {{program}} will not include activity on
+     the :term:`secondaries <secondary>`.
+     If the :term:`primary` is not reachable, {{program}} returns an
+     error message. See:
+     :ref:`Use with Replica Sets <mongotop-use-replicasets>`
+     for more information.
 ---
 program: mongotop
 name: port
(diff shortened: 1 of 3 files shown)


source/reference/program/mongotop.txt
```diff
@@ -174,22 +176,27 @@ This command produces the following output:
...
+                       ns    total    read    write          2014-12-19T15:47:01-05:00
+       admin.system.roles      0ms     0ms      0ms
+     admin.system.version      0ms     0ms      0ms
+                 local.me      0ms     0ms      0ms
+           local.oplog.rs      0ms     0ms      0ms
+   local.replset.minvalid      0ms     0ms      0ms
+        local.startup_log      0ms     0ms      0ms
+     local.system.indexes      0ms     0ms      0ms
+  local.system.namespaces      0ms     0ms      0ms
+     local.system.replset      0ms     0ms      0ms
```
tychoish: local.system.indexes and local.system.namespaces should only appear in mmapv1? either want to note that or produce an alternate output example?

source/includes/options-mongotop.yaml
```diff
@@ -128,8 +131,11 @@ inherit:
   program: _shared
   file: options-shared.yaml
 post: |
-  {{role}} is required for :program:`mongod`
-  and :program:`mongos` instances that use :ref:`authentication`.
+
+
```
tychoish: extra whitespace

source/reference/program/mongotop.txt
schmalliso: I've added a note!

source/includes/options-mongotop.yaml
schmalliso: Required for the `versionchanged` to render properly!

schmalliso:
Merged.

PR #2102, opened 2015-01-10
Update fact-installation-directories.rst

sclaughl:
Made 'user' singular as it should be.


sclaughl:
Extremely trivial change that makes installation docs easier to read and understand.

kay-kim:
Merged.  Thank you.

PR #2103, opened 2015-01-12
DOCS-4586 - correct /etc/mongodb.conf to /etc/mongod.conf

ehershey:


tychoish:
merged and backported to 2.6

PR #2107, opened 2015-01-15
aggregation $sort clarification

schmalliso:
- Clarifies that $sort cannot use the index if $group, $unwind, or
  $project occur before $sort.

- Should probably be backported to v2.6


tychoish:
merged

PR #2110, opened 2015-01-16
Update capped-collections.txt

mcascallares:
Removal of 'delete operations' from intro wording since delete operation is not supported in capped collections


tychoish:
merged

PR #2121, opened 2015-01-21
Update configure-sharded-cluster-balancer.txt

visualzhou:
_secondaryThrottle is true by default in 2.8, which is equivalent to { w: 2 }.


kay-kim:
Merged.  Thank you!

PR #2122, opened 2015-01-22
Via + With cannot go together

FrancescaK:
Removed "with" from "You can install MongoDB with via ``brew`` with several different options. Use"


tychoish:
merged

PR #2123, opened 2015-01-22
WRITING-913 feedback

mpaik:


tychoish:
merged.

in the future only mention public tickets in commit messages.

PR #2125, opened 2015-01-22
WRITING-906 - projection/elemMatch clarification

mpaik:


source/reference/operator/projection/elemMatch.txt
```diff
@@ -54,7 +54,7 @@ assumes a collection ``school`` with the following documents:
                  { name: "ajax", school: 100, age: 7 },
                  { name: "achilles", school: 100, age: 8 },
               ]
-   }
+   },
```
tychoish: these commas aren't needed, and are kind of confusing because this isn't an array.

tychoish:
merged

PR #2136, opened 2015-01-23
notes journal writes are atomic

schmalliso:


tychoish:
merged and backported.

PR #2139, opened 2015-01-24
Fixed $strcasecmp comparison to match output

LarryBattle:
`13q4` should have been used, not `13q3`.


kay-kim:
This has been merged.  Thank you.

Kay Kim

PR #2140, opened 2015-01-26
Update near.txt

nevernet:
by follow https://jira.mongodb.org/browse/SERVER-13218, we should place $minDistance and $maxDistance inside $geometry object.

### DIFF
--- a/source/reference/operator/query/near.txt
+++ b/source/reference/operator/query/near.txt
@@ -31,10 +31,10 @@ Definition
         $near: {
            $geometry: {
               type: "Point" ,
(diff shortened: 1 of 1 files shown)


kay-kim:
Hi -- 
thank you so much for your pull request.  I believe the syntax has changed a bit further to have it be inside the $near operator, not necessarily $geometry object; however, having the $min/$maxDistance inside the $geometry will not error.

kay-kim:
Sorry - accidental tab and return while typing -- oops.

Anyhow, as such, will close the pull request, but we definitely appreciate your time and work. Much  thanks.

Regards,

Kay

PR #2141, opened 2015-01-29
Update URI docs to include `.collection` information.

JasonGiedymin:
Update URI docs to include `.collection` information.


mpaik:

PR #2142, opened 2015-01-30
Update perform-two-phase-commits.txt

rajeshwarpatlolla:
It is bit confusing, that whether the quote(") has opened or closed.
So i modified from opening quote to closing quote.


rajeshwarpatlolla:
Updated with next pull request (#2142).

PR #2147, opened 2015-02-02
DOCS-4663: include with token -> extract

schmalliso:


source/reference/operator/aggregation/cmp.txt
```diff
@@ -1,4 +1,4 @@
-==================
+e==================
```
tychoish: eep

schmalliso: done.

tychoish:
merged

PR #2149, opened 2015-02-03
Update apiargs-option-connection-string-w-param.yaml

jyemin:
Removed -1 option for w, as drivers have either deprecated it or never supported it in the first place.


tychoish:
merged

PR #2154, opened 2015-02-05
Update replica-set-elections.txt

ramsawroop420:
added except Priority 0 members as priority 0 cannot trigger an election and the line "In the default, all members also can trigger an election" seems to suggest otherwise. Hope this is correct understanding.  Do review and edit if i am right about this..

### DIFF
--- a/source/core/replica-set-elections.txt
+++ b/source/core/replica-set-elections.txt
@@ -153,7 +153,7 @@ eligibility to become a :term:`primary`. In an election, the replica
 set elects an eligible member with the highest
 :data:`~local.system.replset.members[n].priority` value as primary. By
 default, all members have a priority of ``1`` and have an equal chance
-of becoming primary. In the default, all members also can trigger an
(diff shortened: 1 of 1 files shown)


kay-kim:
Hi -- Thank you for your pull request.  I believe, however, that the current sentence is correct.  By default, members have priority of 1, and as such, they can trigger an election since no member by default is priority 0.  As such, will close this pull request; however, will update that section to be more clear.

Thanks again for your efforts!  

Regards

ramsawroop420:
Thanks for correcting it. Can you suggest best forum where i can clarify such doubts. Being a begineer it is very important to understand the basics..

kay-kim:
People tend to use stackoverflow or sometimes open a JIRA ticket if they believe there is a mistake in the docs or if the docs could be clearer or what you did (do a pull request with a suggested fix) is fine as well.

ramsawroop420:
Thank

PR #2158, opened 2015-02-09
Fix grammar in create-tailable-cursor.txt

jmikola:


tychoish:
merged and backported

PR #2162, opened 2015-02-11
Update query-optimization.txt

kerneljake:
The word "returned" implies that you can workaround this limitation using a projection, but in fact you cannot (per SERVER-2104).

### DIFF
--- a/source/core/query-optimization.txt
+++ b/source/core/query-optimization.txt
@@ -144,7 +144,7 @@ An index **cannot** cover a query if:
   a :ref:`multi-key index <index-type-multikey>` index and cannot
   support a covered query.
 
-- any of the returned indexed fields are fields in subdocuments.
+- any of the indexed fields are fields in subdocuments.
   [#index-embedded-document-fields]_ For example, consider a collection
   ``users`` with documents of the following form:
 


kay-kim:
Hi -- isn't the current page and the SERVER-2104 stating that if you have an index 
   db.foo.ensureIndex( { a: 1, "b.x": 1 } )
then
  db.foo.find( { a: { $lt: 1 } }, { a: 1, _id: 0 } ) will be covered
whereas
  db.foo.find( { a: { $lt:  1} }, { a: 1, "b.x": 1, _id: 0 } ) will not be covered

kerneljake:
I agree with the above example, buy my example is different.
Below is an example of a literal interpretation of the sentence in an attempt to get around this limitation, but it doesn't work.

{code}
> db.foo.insert({ tenant:1, user:{ name:"tester", password:"secret"}})

> db.foo.ensureIndex({tenant:1, "user.name":1})

> db.foo.find({tenant:1, "user.name":"tester"}, {user:0, _id:0}).explain(true).executionStats.totalDocsExamined
1
{code}

Since the indexed field in the subdocument is not returned, I would expect the query to be covered, but it is not covered.  The phrase "any of the returned index fields are fields in subdocuments" implies that you can project away a subdocument containing an indexed field and still achieve a covered query.

kay-kim:
ah -- I see.  How about "If any of the indexed field in the query predicate or returned in the projection ..."
to cover both examples.

kerneljake:
Yes, that sounds good!

kay-kim:
merged with changes as discussed.  thank you!

PR #2164, opened 2015-02-11
Explain what --oplogReplay really does.

drmirror:
This information is needed if you want to replay an arbitrary oplog, not just an oplog slice that was generated by mongodump --oplog.


drmirror:
Oops, that should have been "To find the oplog..." in the beginning, rather than "The find the oplog..."

kay-kim:
Thanks Andre!  This has been merged.

PR #2165, opened 2015-02-12
Copy edit for typo in file path

epc:


kay-kim:
thanks ed.  was already fixed as part of another commit.

PR #2173, opened 2015-02-17
fix `.. versionchanged` directive syntax

rutsky:


kay-kim:
Thank you for your fix.  This has been merged.

Regards,

Kay Kim

PR #2176, opened 2015-02-24
DOCS-4753: adds/updates mongostat columns for WT

schmalliso:
LGTM'd by Mike O'Brien.

### DIFF
--- a/source/reference/program/mongostat.txt
+++ b/source/reference/program/mongostat.txt
@@ -148,10 +148,33 @@ average operations per second.
    ``local|replicated`` commands.
 
 .. describe:: flushes
+   
+   .. versionchanged:: 3.0.0
+      When using :ref:`storage-wiredtiger`, ``flushes`` refers to the
+      number of wiredTiger checkpoints triggered between each
+      polling interval.
+      
+      With :ref:`storage-mmapv1`, ``flushes`` represents the number of
+      :term:`fsync` operations per second.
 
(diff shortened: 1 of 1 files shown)


source/reference/program/mongostat.txt
```diff
@@ -148,11 +148,34 @@ average operations per second.
...
 
 .. describe:: flushes
+   
+   .. versionchanged:: 3.0.0
+      When using :ref:`storage-wiredtiger`, ``flushes`` refers to the
+      number of wiredTiger checkpoints triggered between each
+      polling interval.
+      
+      With :ref:`storage-mmapv1`, ``flushes`` represents the number of
+      :term:`fsync` operations per second.
```
kay-kim: could we make this a bit parallel with the leading prepositions?

source/reference/program/mongostat.txt
```diff
@@ -148,11 +148,34 @@ average operations per second.
...
 
-   The number of :term:`fsync` operations per second.
+.. describe:: dirty
+
+   .. versionadded:: 3.0.0
+
+   Only for :ref:`storage-wiredtiger`. The percentage of the wiredTiger
```
kay-kim: While we haven't decided upon consistent use for  ``wiredTiger`` vs WiredTiger, we don't use just wiredTiger. I think we should go with the general WiredTiger.  ditto with MMAPv1

kay-kim:
merged.

PR #2200, opened 2015-03-04
DOCS-4913 - redact meta driver

mpaik:


mpaik:
Merged with whitespace edit

PR #2204, opened 2015-03-05
Minor fix for missing parenthesis in runCommand

gianpaj:
All the other Query Plan Cache commands look good:
http://docs.mongodb.org/manual/reference/command/nav-plan-cache/


kay-kim:
merged!

PR #2209, opened 2015-03-09
DOCS-4568: updates to extended json page

schmalliso:
- clarifies that you need quotes around NumberLong values or
  risk loosing accuracy (numbers can become negative)
- updates undefined to clarify that it cannot be used in a query
  except by searching using the $type operator

**Should probably be back-ported to 2.6**

### DIFF
--- a/source/reference/mongodb-extended-json.txt
+++ b/source/reference/mongodb-extended-json.txt
@@ -294,6 +294,30 @@ Undefined Type
 
    The representation for the JavaScript/BSON undefined type.
 
+   You *cannot* use ``undefined`` in query documents.
+   Consider the following document inserted into the ``people`` collection:
+   
+   .. code-block:: sh
+   
+      db.people.insert( { name : "Sally", age : undefined } )
+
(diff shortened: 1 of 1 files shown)


source/reference/mongodb-extended-json.txt
```diff
@@ -294,6 +294,30 @@ Undefined Type
 
    The representation for the JavaScript/BSON undefined type.
 
+   You *cannot* use ``undefined`` in query documents.
+   Consider the following document inserted into the ``people`` collection:
+   
+   .. code-block:: sh
+   
+      db.people.insert( { name : Sally, age : undefined } )
```
kay-kim: that will fail.  Sally is a string

schmalliso: TIL: quotation marks are optional on key names, but not on values. 

I quote everything, normally (keys and values). Will update.

schmalliso:
Merged to master and also backported to 2.6.

PR #2217, opened 2015-03-11
Updated examples to use aggregation cursor and crud spec for 2.0

christkv:


primer/source/includes/examples-remove.yaml
```diff
@@ -208,7 +207,7 @@ operation:
     language: javascript
     code: |
       var removeRestaurants = function(db, callback) {
-         db.collection('restaurants').remove( {}, function(err, results) {
+         db.collection('restaurants').removeMany( {}, function(err, results) {
```
kay-kim: Hey Christian -- 
this example, it's not deleteMany?  So, in node, we have deleteOne, deleteMany, and removeMany?

kay-kim:
merged.  also updated removeMany to deleteMany in that example

PR #2221, opened 2015-03-13
Visually highlight the warning about using $min xor $max

devkev:
Following up from DOCS-4015, the second paragraph in the "$min without $max" and "$max within $min" sections should be more noticable, since it is warning the user of hazardous usage that is best avoided.

If it's not appropriate to have such a warning box inside a "fact" file (rather than a "warning" file), then it should be easy enough to rearrange as necessary.


mpaik:
edited and merged

PR #2225, opened 2015-03-16
Note that mongod --shutdown not available in latest version

jcrben:
Per http://stackoverflow.com/questions/18835603/mongodb-shutdown-option-unavailable this option has been unavailable since at least v2.4.5.

Maybe we're missing something but I get the same message. If it is truly removed, it seems reasonable to remove the text entirely and I can update the pull request accordingly.

### DIFF
--- a/source/tutorial/manage-mongodb-processes.txt
+++ b/source/tutorial/manage-mongodb-processes.txt
@@ -139,6 +139,7 @@ authentication enabled.
 Use ``--shutdown``
 ~~~~~~~~~~~~~~~~~~
 
+Note: this is not available in the latest version.
 From the Linux command line, shut down the :program:`mongod` using the
 :option:`--shutdown <mongod --shutdown>` option in the following command:
 


mpaik:
--shutdown works fine in mongod 3.0.0. Cannot reproduce.

jcrben:
Any idea why so many people are getting "error command line: unknown option shutdown
use --help for help" when they try running this?

There's a few hundred views and 2 votes for an answer which says: "Since the "--shutdown" option is not available in newer versions of mongo, you now have to kill the process manually"

schmalliso:
@jcrben What OS are you using? I'm able to reproduce your issues on OS X, and `--shutdown` doesn't appear in my `--help output`, while @mpaik is on a Linux machine, and does have it.

mpaik:
This definitely works on Linux, and [the `mongod` documentation page](http://docs.mongodb.org/manual/reference/program/mongod/#cmdoption--shutdown) on all published manual versions from 2.2 through 3.0 clearly indicate that `--shutdown` is only available on Linux. [The page](http://docs.mongodb.org/manual/tutorial/manage-mongodb-processes/#use-shutdown) the pull request is against also says "From the Linux command line" on all published manual versions from 2.4 through 3.0, though an ambiguity exists in 2.2. It is possible that an older version of the manual had a bug, but the documentation is currently correct as written, as far as I can tell.

jcrben:
Yeah, I'm using OS X. I guess I glossed over that Linux bit after I saw the Stack Overflow page. My bad. I'll clarify this on the Stack Overflow thread, but I wonder if there should be a clarification in the tutorial page. Many new developers first encounter mongo when developing locally on an OS X machine, and new developers are the ones using the tutorial.

PR #2234, opened 2015-03-19
Update python code samples for PyMongo 3.0.

behackett:
Also includes edits to fix pep8 issues.


kay-kim:
merged.  Thank you so much!

PR #2235, opened 2015-03-21
fixed secondary throttle section

asya999:
Clarified writeConcern, added missing space, removed incorrect braces.


kay-kim:
merged!

PR #2238, opened 2015-03-23
DOCS-1891: clarifies 'reasonable' pool sizes

schmalliso:
Adds note about what a good starting point for the 'reasonable pool size' is, as specified in the Production Checklist.


kay-kim:
merged

PR #2240, opened 2015-03-23
DOCS-4809: updates behavior for -j alias and adds stopOnError to mongorestore

schmalliso:
Also pulls out note on mongoimport that indicates that `--stopOnError` doesn't stop for some types of errors, as that is no longer true.


kay-kim:
merged

PR #2245, opened 2015-03-26
make it clearer we're talking about two different files

bgrabar:
(the two strings are long and differ by only one character,
which can be confusing)


bgrabar:
Why was this closed?

PR #2248, opened 2015-03-27
DOCS-3547: Document MongoDB uninstallation

i80and:


kay-kim:
merged.  tweaked a little when backporting to v2.6:
-- 2.6 -- suse only on enterprise
         -- we seem to state that the data directory in existing doc is /data/db

PR #2250, opened 2015-03-30
DOCS-5125 Fix syntax for maxTimeMS

NoOutlet:
Description of maxTimeMS included an alternative to the
shortcut which used incorrect syntax. The incorrect command
did not limit the time for retrieving results and instead
created a inclusive projection on a field named '$maxTimeMS'
which would hopefully not be an actual field in the user
database.

### DIFF
--- a/source/reference/operator/meta/maxTimeMS.txt
+++ b/source/reference/operator/meta/maxTimeMS.txt
@@ -22,7 +22,7 @@ $maxTimeMS
 
    .. code-block:: javascript
 
-      db.collection.find( { } , { $maxTimeMS: 100 } )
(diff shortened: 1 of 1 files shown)


kay-kim:
Hi -- thank you very much for the fix.  You are absolutely correct.  Before we can merge this fix in, could you sign the http://www.mongodb.com/legal/contributor-agreement?  Thanks again for the fix.

Regards,

Kay Kim

NoOutlet:
Hi Kay Kim,

I did sign it just before submitting the pull request. I'm not sure how long that takes to process though.

Thanks!
Jon

kay-kim:
Ah -- it takes 24 hours.  Thanks for signing : )  Merged in your changes.  Thanks again.

PR #2255, opened 2015-03-30
DOCS-5120: updates --password behavior to reflect new reality

schmalliso:


schmalliso:
Brief panic double-checking other references to `--password`, but they are all to its use with `mongod`/`mongo`, which doesn't apply here.

kay-kim:
merged

PR #2263, opened 2015-04-07
DOCS-5169: Fix missing brace in $push, rephrase intro in $addToSet

i80and:


source/reference/operator/update/addToSet.txt
```diff
@@ -9,10 +9,8 @@ Definition
 
 .. update:: $addToSet
 
-   The :update:`$addToSet` operator adds a value to an array only *if* the
-   value is *not* already in the array. If the value *is* in the
-   array, :update:`$addToSet` does not modify the
-   array.
+   The :update:`$addToSet` operator adds a value to an array unless the value
+   is already present, in which case :update:`$addToSet` does nothing.
```
kay-kim: This is fine, but I believe we want to clarify that it's only the array with the dupe that $addToSet does nothing -- as in, it's not all or nothing given the the operator can work on multiple fields.

db.foo.update( { }, { $addToSet: { x: 1, y: 2 } } )

kay-kim:
merged

PR #2264, opened 2015-04-07
DOCS-5171 & DOCS-5172: mongostat stat corrections

schmalliso:
Note: The steps thing fixed a build error I was getting. Do with it what you will.

### DIFF
--- a/source/reference/program/mongostat.txt
+++ b/source/reference/program/mongostat.txt
@@ -179,15 +179,21 @@ average operations per second.
    megabytes. This is the total data size at the time of the last
    :program:`mongostat` call.
 
-.. describe:: size
+.. describe:: vsize
 
    The amount of virtual memory in megabytes used by the process at
    the time of the last :program:`mongostat` call.
 
 .. describe:: non-mapped
 
-   The total amount of virtual memory excluding all mapped memory at
+   .. versionchanged:: 3.0.0 
+   Only for :ref:`storage-mmapv1`.
+
+   *Optional*. The total amount of virtual memory excluding all mapped memory at
    the time of the last :program:`mongostat` call.
+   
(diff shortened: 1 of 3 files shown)


source/reference/program/mongostat.txt
```diff
@@ -179,15 +179,21 @@ average operations per second.
...
+.. describe:: vsize
 
    The amount of virtual memory in megabytes used by the process at
    the time of the last :program:`mongostat` call.
 
 .. describe:: non-mapped
 
-   The total amount of virtual memory excluding all mapped memory at
+   .. versionchanged:: 3.0.0 
+   Only for :ref:`storage-mmapv1`.
```
kay-kim: does this compile correctly without the blank line in between versionchanged and the next line?

schmalliso: Yes ma'am.

non-mapped

kay-kim: so weird.  I can't even pull this pull request.  if I do a fake
.. versionchanged:: 3.0.0
This is a test

Mine nags at me, but we seem to have build errors due to michael's debian step and one of the toc's is looking for the core/tag-sharding in  toc-sharded-cluster-data.yaml which doesn't allow for any actual builds.  eh.

schmalliso: I fixed the Debian step here, since it wasn't building at all for me either.

As for tag-aware-sharding, I moved it, but it should be fine (https://github.com/mongodb/docs/commit/6d9e8f2acbab065aa8244776b47614a9fc782970)? I just copy-pasted it across. Maybe I'll cherry-pick this commit into a new master.

schmalliso:
Perhaps now it will be more content?

PR #2267, opened 2015-04-08
DOCS-4974: removes reference to reading datafiles directly

schmalliso:
Also moves the mongodump-behavior reference to above the Use header, rather than below it, because that was weird.


kay-kim:
merged

PR #2269, opened 2015-04-09
replSetStepDown: it's secondaryCatchUpPeriodSecs, not secondaryCatchupPeriodSecs

devkev:
This might seem minor, but the effect is that after copy-pasting the option, it apparently has no effect on the command.  This makes it seem like the option is broken in the server, when in fact it's just a small typo in the docs.


kay-kim:
merged!  thank you.

PR #2271, opened 2015-04-09
compression clarifications

kgorman:
Based on my conversations with Asya about compression, I made a couple edits that hopefully clarify a bit what each attribute in means specifically taking into account compression.


kay-kim:
Thank you for your commit!  We've merged with edits.

Regards,

Kay Kim

PR #2275, opened 2015-04-20
improvement on materialized path tree case

OthmanAba:


mpaik:
Using alternate fix. Thanks for your contribution!

PR #2282, opened 2015-04-24
Update query-optimization.txt

mgorgei:
Comma usage was confusing before removing them.

Specifically, you do not separate two related items by a comma.  I don't know a rule for the other one, but imagine 'ascending index' without the series, it's an unnatural pause.


kay-kim:
Hi.  Thank you for your change.   It has been merged!  I believe the original usage of the commas were more in a parenthetical sense, like "add an ascending (or a descending) index".

Thanks again!

Regards,

Kay Kim

PR #2284, opened 2015-04-28
Grammar fix in MongoDB for Windows

dandv:


kay-kim:
Hey Dan -- I think you forgot to actually save the file with the fix.

I'll go ahead and close this, but will change "does not" to "do not" in the sentence.  

The 64-bit versions of MongoDB do not work with 32-bit Windows.

Thanks as always. 

Regards,
Kay Kim

dandv:
That is odd. Thanks Kay!

PR #2286, opened 2015-04-29
Update TCP keepalive commands for Linux

jamestyj:


kay-kim:
Hey James -- 
we're not ignoring you.  The include files are also used in the production notes under the MongoDB on Virtual Env and we have a request to update that section, so will incorporate pull in your merge when we update that section in one fell swoop.

jamestyj:
No problem and thanks for the update.

kay-kim:
Merged into the local branch for keeaplive work.  I've added you to code review -- lucky, lucky you.

PR #2289, opened 2015-05-04
Update configure-a-hidden-replica-set-member.txt

kerneljake:
Make voting language consistent for hidden members along the lines of DOCS-4722.


kay-kim:
merged.  thank you.

PR #2292, opened 2015-05-07
DOCS-5284: improves upsert/upsertFields wording

schmalliso:
@kay-kim I've made the tweaks you suggested. I think things are clearer now :smiley_cat:


kay-kim:
thank you.  merged.

PR #2298, opened 2015-05-19
DOCS-5453: 2.6.10 releease notes

tychoish:


kay-kim:
merged -- just an fyi, somehow, your commit looks as though I made the commit and you merged it to your branch, and then I merged that in.  eh.

PR #2302, opened 2015-05-25
Fix date format and ObjectIds

gianpaj:
The ObjectIds order is based the order of documents inserted and the sort()


kay-kim:
merged.  thank you!

PR #2303, opened 2015-05-27
Update production-notes.txt

artickl:
Received warning:
*\* WARNING: You are running on a NUMA machine.
*\*          We suggest launching mongod like this to avoid performance problems:
*\*              numactl --interleave=all mongod [other options]

After installation mongod 3.0.3 on Ubuntu 14.04. After checking how to enable it - found that upstart script (/etc/init/mongod.conf) taken care of it themself if package "numactl" is installed.

So, "apt-get install numactl && service mongod restart" fixed this warning, but I spend 10-20 minutes to dig it in... So, would like to save time to somebody else =)

---

This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/2303)

### DIFF
--- a/source/administration/production-notes.txt
+++ b/source/administration/production-notes.txt
@@ -298,6 +298,11 @@ To fully disable NUMA behavior, you must perform both operations. For more
 information, see the `Documentation for /proc/sys/vm/*
 <http://www.kernel.org/doc/Documentation/sysctl/vm.txt>`_.
 
+Configuring NUMA on Ubunutu
+```````````````````````````
+On Ubuntu, simply install "numactl" package and restart mongod. Upstart script will automatically 
+start mongo with correct parameters.
+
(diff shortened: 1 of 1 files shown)


p-mongo:
I would think that for a *production* deployment, if one is running on a numa machine they would be well served by reviewing the settings and ensuring the settings are correct, rather than relying on unknown defaults. Hence I would suggest not making the change proposed in this PR.

kay-kim:
Hi @artickl  -- 
thank you much for this pull request and apologies for the long delay.  Recently we have updated the section
https://docs.mongodb.com/manual/administration/production-notes/#configuring-numa-on-linux

and as such, will close this pull request without merge.  However, we very much appreciate your taking the time to help us improve the docs and apologies once again for the long delay.

Regards,

Kay Kim

PR #2306, opened 2015-05-29
DOCS-5511: Add Windows 2012 R2 to the Production list

markbenvenuto:
Add Windows 2012 R2 to the Production list for supported operating systems.


kay-kim:
merged and published.  thank you.

PR #2309, opened 2015-06-11
DOCS-5601: 3.0.4-rc0 release notes

tychoish:
Should probably update the version from 3.0.3 to 3.04-rc0 in the version
config, and then change the dates and version when GA goes out.


kay-kim:
merged.

PR #2312, opened 2015-06-15
Chunks can still migrate from a maxSize shard

devkev:
It is untrue that a shard which is over its maxSize value will _not_ have chunks migrated away from it.

This is correctly stated in the docs for the maxSize parameter for [addShard](http://docs.mongodb.org/manual/reference/command/addShard/), which says:

> The maxSize constraint prevents the balancer from migrating chunks to the shard when the value of mem.mapped exceeds the value of maxSize.

It is also easy to check empirically: start a 2 shard cluster, shard a collection, add `maxSize: 1` to the primary shard, then split once so there are 2 chunks, and observe that the balancer still migrates a chunk away from the primary (so that it's now 1 chunk on each shard).  If you manually move both chunks onto the _non-maxSize_ shard, the balancer will not move a chunk onto the maxSize shard (which is correct).

---
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/2312)

### DIFF
--- a/source/tutorial/configure-sharded-cluster-balancer.txt
+++ b/source/tutorial/configure-sharded-cluster-balancer.txt
@@ -57,9 +57,9 @@ The ``maxSize`` field in the :data:`~config.shards` collection in the
 :ref:`config database <config-database>` sets the maximum size for a
 shard, allowing you to control whether the balancer will migrate chunks
 to a shard. If :data:`~serverStatus.mem.mapped` size [#local-limitation]_ is above a shard's
-``maxSize``, the balancer will not move chunks to the shard. Also, the
-balancer will not move chunks off an overloaded shard. This must happen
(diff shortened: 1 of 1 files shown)


p-mongo:
The documentation that this PR changes has been rewritten, and currently reads:

<pre>
Change the Maximum Storage Size for a Given Shard
-------------------------------------------------

By default shards have no constraints in storage size. However, you can set a
maximum storage size for a given shard in the sharded cluster. When 
selecting potential destination shards, the balancer ignores shards 
where a migration would exceed the configured maximum storage size.

</pre>

I think this PR can be closed since the erroneous information is no longer present.

kay-kim:
Hey Kevin -- 
closing as Oleg correctly points out, we've update the page.  Sorry for the delay in housekeeping.  Much appreciated as always.  Kay

PR #2313, opened 2015-06-17
Update connectionStatus.txt

mcascallares:
it should be connection instead of collection


kay-kim:
Thank you!  This has been merged and backported.

PR #2326, opened 2015-06-26
Fix various warnings

i80and:
* Remove meta/includes.txt; we don't use it, and it clutters the log with an error and a warning
* Fix broken link in the 2.2 release notes


kay-kim:
merged

PR #2328, opened 2015-06-26
Fix links

i80and:


kay-kim:
merged.  this probably fixes DOCS-5693.

PR #2333, opened 2015-07-07
Minor documentation fix

vaultah:
"is" was missing and since the prefix is a required part of the URI, "host" isn't the only required part.

### DIFF
--- a/source/reference/connection-string.txt
+++ b/source/reference/connection-string.txt
@@ -44,7 +44,7 @@ The components of this string are:
 
 #. ``host1``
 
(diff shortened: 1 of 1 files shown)


kay-kim:
Hi -- Much apologies for the delay in response and thank you so very much for finding the error as well as providing the fix.  Recently, we had gone through and updated the page.  As such, we did reduce that whole problematic sentence to just "Required."  As such, will go ahead and close this pull request.  But thanks again for taking the time to do this.  It's very much appreciated, and sorry again for the delay in response. 

Warmest regards,

Kay Kim

PR #2344, opened 2015-07-27
remove $ from minDistance

asya999:


kay-kim:
merged!  Thank you and will republish later this afternoon with some additional changes (such as updated examples for doc validation).

PR #2348, opened 2015-07-28
DOCS-5724: Add missing cyrus-sasl-plain and cyrus-sasl-gssapi deps

i80and:


kay-kim:
merged to both 2.6 and 3.0

PR #2351, opened 2015-07-30
`zypper addrepo` requires sudo

devkev:
The docs for enterprise (steps-install-mongodb-enterprise-on-suse.yaml) correctly state this.


kay-kim:
merged!  thank you.

PR #2352, opened 2015-07-30
Non-interactive SUSE installation

devkev:
Installation instructions for other distros are non-interactive, so the SUSE ones should be also.  For zypper, this is `-n` aka `--non-interactive` (*before* the `install` sub-command).


kay-kim:
merged!  thank you.

PR #2356, opened 2015-08-06
DOCS-5246: adds reference page for setVerboseShell() method

schmalliso:
This will either need redirects or to be backported to 2.4 / 2.6, if it existed in 2.4 (which I think it did?).


kay-kim:
merged!

PR #2359, opened 2015-08-07
Elevate the Getting Started section

i80and:


source/index.txt
```diff
@@ -16,6 +16,37 @@ introduces key concepts in MongoDB, presents the query language, and
...
+
+       :gettingstarted:`C# Edition </csharp>`
+
+     -
+
+Topics
+------
+
+The MongoDB manual is under constant development. See the :doc:`/about`
+for more information on the MongoDB Documentation project.
```
kay-kim: You might want to get rid of these sentences and clarify that the list of topics here is not exhaustive.

kay-kim:
merged

PR #2360, opened 2015-08-09
the link to Python is missing /

asya999:
link was going to pythoncurrent so the slash wasn't showing up (wrong slash?)


kay-kim:
merged. thank you.

PR #2365, opened 2015-08-15
Fix broken repo URL

dandv:
See http://stackoverflow.com/questions/28945921/e-unable-to-locate-package-mongodb-org#comment47314415_28966356


kay-kim:
Hi Dan -- 
thanks as always for your pull request.  The expectation with using $(lsb_release -sc) is that the user's ubuntu's release version would resolve to either trusty or precise.  Instead of merging in this pull, I'll open a ticket to note that the command expects the lsb_release to return one of the supported versions.

Regards,

Kay Kim

PR #2366, opened 2015-08-17
production-notes: tweak Linux NUMA disable

devkev:
1. Swap the order of the `numactl` and `zone_reclaim_mode` commands. This is better because if a user manually follows the old steps, they may run `numactl --interleave=all mongod` *before* fixing `zone_reclaim_mode`, and thereby still receive `zone_reclaim_mode` startup warnings.

2. Adjust the `zone_reclaim_mode` command to be an appropriate sudo-style command (simply prepending `sudo` won't work), indicating that the command needs to be run as root (and how).  Also provide a more modern `sysctl` based variant (again with sudo).

3. Add `<options>` to the `numactl` command to indicate that these can be specified if desired.

4. "may" -> "should"

5. "including ..., or clients" -> "including ..., and clients"

### DIFF
--- a/source/administration/production-notes.txt
+++ b/source/administration/production-notes.txt
@@ -284,21 +284,28 @@ Please consult your system documentation for details.
 Configuring NUMA on Linux
 `````````````````````````
 
-When running MongoDB on Linux, you may instead use the ``numactl``
-command and start the MongoDB programs (:program:`mongod`, including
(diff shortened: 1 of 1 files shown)


schmalliso:
Squashed in with https://github.com/mongodb/docs/pull/2568. Sorry for the delay :(

PR #2368, opened 2015-08-18
production-notes: tweak WT XFS language

devkev:
I found the old phrasing slightly ambiguous to parse, and prefer this way.


kay-kim:
merged.  thank you.

PR #2373, opened 2015-08-31
DOCS-6108: Make Ubuntu installation instructions more explicit

i80and:
Deliberately left the enterprise "specific branch" instructions as-is; they're more "power-user" oriented, and I didn't want to clutter the step.


kay-kim:
merged

PR #2375, opened 2015-09-06
explanation did not match example.

ardis:


kay-kim:
Merged!  Thank you very much.

Regards,

Kay Kim

PR #2377, opened 2015-09-11
Small wording fix

razzius:


kay-kim:
Hello --
Thank you for this pull request and apologies in the lateness of this response.  The file you have modified is auto-generated from another file, and as such, I will close this pull request without merging, but will make a ticket to ensure that the other file is updated accordingly.

Thanks again for taking the time to make the change and the pull.  

Regards,

Kay Kim

PR #2378, opened 2015-09-12
Typo

marekgoczol:
through**o**ut instead of through**p**ut


kay-kim:
Hi -- thank you for taking the time to make and submit your changes.  However, in this case, the correct word is throughput in this context where we're talking about the amount of writes and reads being processed by the system.

Once again, we appreciate your time and efforts.  

Regards, 

Kay Kim

PR #2380, opened 2015-09-16
FATAL is no longer a valid state in 3.0

aksaharan:
FATAL is no longer a valid state in 3.0. It does not need to be defined here since it is not mentioned as a valid state at the top in this page.

---
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/2380)

### DIFF
--- a/source/reference/replica-states.txt
+++ b/source/reference/replica-states.txt
@@ -111,8 +111,3 @@ Members in any error state can't vote.
    members. In this case, the old primary member reverts those writes. During
    :doc:`rollback </core/replica-set-rollbacks>`, the member will have
    :replstate:`ROLLBACK` state.
-
-.. replstate:: FATAL
-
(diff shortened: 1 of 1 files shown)


kay-kim:
Hey @aksaharan  -- you are absolutely correct and sorry for the long delay.  Not quite sure why it wasn't done as part of https://jira.mongodb.org/browse/DOCS-4333.   Because of the delay on our part, we can't rebase and merge because the file has changed a bit.  But, what I'll do is to just delete that state from the branches (and backport) and close this pull.

Sorry about that.  

Regards,

Kay

kay-kim:
Oh, actually, we just annotated it as _Removed in v3.0_  about 4 years ago.

But I am guessing at this point that removing it wholesale would be good for the 3.4+ versions of the docs.

PR #2385, opened 2015-09-28
DOCS-4968: findAndModify return document

i80and:


kay-kim:
merged and out of your hair. Will publish this today.

PR #2397, opened 2015-10-14
Fix GPG key reference in Debian/Ubuntu install instructions.

breser:
This was updated in the Verify Integrity of MongoDB Packages but not
here.  Without this fix the installation instructions fail.

### DIFF
--- a/source/includes/steps-install-mongodb-on-debian.yaml
+++ b/source/includes/steps-install-mongodb-on-debian.yaml
@@ -7,7 +7,7 @@ action:
     <https://docs.mongodb.org/10gen-gpg-key.asc>`_ to the system key ring.
   language: sh
   code: |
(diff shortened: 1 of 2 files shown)


breser:
And now they're back to the being signed with 7F0CEB10.  So perhaps the packages were unintentionally signed with the wrong key and you've fixed them.  I suppose changing them to the release package presents a problem for anyone who's already installed the repo this way because they'll have to know to add the additional key.  But it'd be nice if this discrepancy between signing with a project key in some places and signing with an individuals key in others was called out and explicitly confirmed as valid.

i80and:
Thanks for the report, Ben!

We're currently in a transition regarding our signing process. We may switch to signing 3.0 packages with 24F3C978 (the project key) in the future, but for now, 7F0CEB10 is the right key.

Agreed that it's a little more confusing than it should be right now, but the situation will be improving!

PR #2399, opened 2015-10-21
removes commands/methods from bypassDocumentvalidation list, copy edits

schmalliso:
Removals from list per Dan's feedback on the larger `bypassDocumentValidation` code review.


kay-kim:
merged!

PR #2401, opened 2015-10-22
Add note about writeConcern for tools being changed

asya999:
writeConcern for ingest for tools is majority as of 3.0


kay-kim:
merged.  thank you.

PR #2404, opened 2015-10-30
DOCS-5882 : aliases with $type

rkumar-mongo:
As per DOCS-5882


kay-kim:
merged into my local master.

PR #2408, opened 2015-11-04
Update security-transport-encryption.txt

kerneljake:
remove duplicate bullet point


kay-kim:
Hi Jake -- thanks so much. I think that was supposed to be  :doc:`/tutorial/configure-ssl` for the first bullet.  I'll close this pull and update the first bullet.  Kay

PR #2412, opened 2015-11-10
DOCS-5960 : explain() now works for distinct() commands

rkumar-mongo:
updated explain page to include new method


kay-kim:
merged

PR #2413, opened 2015-11-10
Partial indexes

schmalliso:
TODO
* It would make sense to add a note to /core/query-optimization about how partial indexes have potentially not-fully-intuitive behavior
* Proof read :smiley_cat:


kay-kim:
merged into local branch

PR #2416, opened 2015-11-19
DOCS-5300: Create supported platforms matrix

i80and:
If we want to backport, please let me know. There's at least one part that does not apply to past releases.


kay-kim:
merged to master and v3.0

PR #2417, opened 2015-11-19
DOCS-6306 - explain().aggregate() functionality clarification

rkumar-mongo:
explain().aggregate() only returns queryPlanner verbosity mode at this time


kay-kim:
merged

PR #2425, opened 2015-11-25
Fix syntax of createCollection with document validation

stbrody:


kay-kim:
Thanks Spencer! This has been merged.

PR #2427, opened 2015-11-30
Adding "or"

dhatcher42:
Should add an "or" between "keyfiles x509 certificates" because they are two separate items.


kay-kim:
Thank you!  This has been merged.

PR #2429, opened 2015-12-02
PLAIN does not work for in-database users in 3.0

Boomtime:
See SERVER-20691, and HELP-1536 for extra context.


kay-kim:
merged!

PR #2430, opened 2015-12-02
Update bitsAnySet.txt

rachnac:
The example should use `35` for bit comparison instead of `50` which is not mentioned.


i80and:
Thanks for the report, @rachnac ! We really appreciate it!

There were a couple other little issues and tweaks that needed to be done in the same area, so I just went around and reworked all the bitwise examples in one go.

PR #2439, opened 2015-12-04
DOCS-6720 remove legacy C++ driver example

ksuarz:
Deletes an example using tailable cursors with the legacy C++ driver, as well as
a link to a third-party blog.


kay-kim:
merged!

PR #2441, opened 2015-12-07
DOCS-6679: Compass documentation

ksuarz:
This adds the documentation for Compass in the MongoDB manual.

There is also an additional commit that handles the layout of the "Additional Products" page, with links to Compass, Ops Manager, and Cloud Manager. This will likely be refactored some time in the future.


kay-kim:
merged!

PR #2444, opened 2015-12-08
DOCS-6620 : getMORE maxTimeMS should specify awaitData behavior

rkumar-mongo:


source/includes/apiargs-dbcommand-getMore-field.yaml
```diff
@@ -32,7 +32,12 @@ position: 4
 name: maxTimeMS
 type: positive integer
 description: |
-  The timeout duration in milliseconds.
+  The time period in milliseconds the :dbcommand:`getMore()` operation will
```
ajdavis: I propose "The maximum time period", otherwise it sounds like the command *will* block waiting for new data. In fact the command returns early if new data arrives before maxTimeMS.

rkumar-mongo: Fixed - thanks for catching that.

ajdavis:
LGTM!

kay-kim:
merged and published

PR #2453, opened 2015-12-09
DOCS-6106: document indexOptionDefaults

ksuarz:
This adds documentation for the `indexOptionDefaults` option for the `create`
command and its shell helper, `db.createCommand()`.


ksuarz:
Let me do one last-minute check that I've properly updated both pages for the command and shell helper before accepting this.

PR #2455, opened 2015-12-09
DOCS-6106: document indexOptionDefaults

ksuarz:
This adds documentation for the `indexOptionDefaults` option for the `create`
command and its shell helper, `db.createCommand()`.


kay-kim:
merged!

PR #2461, opened 2015-12-09
Update bi-connector.txt

kstirman:
The name should be MongoDB Connector for BI.


kay-kim:
merged.  will republish after getting your other pull. Thanks.

PR #2468, opened 2015-12-10
DOCS-6486: readConcern shell helper for find/count

ksuarz:
Adds documentation for the `.readConcern()` helper for the `db.collection.find()`
and `db.collection.count()` methods.

An entry for "read concern" has also been added to the glossary.


kay-kim:
merged

PR #2470, opened 2015-12-11
Add comment about where to connect

justanyone:
All these tutorials should be self-contained and presume no info in previous tutorials.
This means that everywhere we say, "you must update this...", you need to add a 
previous statement for, 'connect to this component'.

Many of customers have sharded replicasets.  Saying 'first connect, then...' is too vague.
Don't presume the user knows which to connect to - primary, replica, arbiter, all 3, config
server, or mongos.  Especially n00bs, don't know which thing to connect to.

### DIFF
--- a/source/tutorial/adjust-replica-set-member-priority.txt
+++ b/source/tutorial/adjust-replica-set-member-priority.txt
@@ -20,7 +20,10 @@ priority.
 Considerations
 --------------
 
-To modify priorities, you update the :rsconf:`members`
+To modify priorities, you can connect to any replicaset member (not a mongos).
+Changes you make there will be picked up by other replica set members.
+
+After connecting, you update the :rsconf:`members`
 array in the replica configuration object. The array index begins with
 ``0``. Do **not** confuse this index value with the value of the replica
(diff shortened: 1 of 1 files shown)


p-mongo:
I believe existing language already covers the proposal made in this PR.

The "procedure" section starts with:

> In the mongo shell, use rs.conf() to retrieve the replica set configuration and assign it to a variable. For example:

The rs.conf is linked to [the shell helper documentation](https://docs.mongodb.com/manual/reference/method/rs.conf/), which is linked to [the server command](https://docs.mongodb.com/manual/reference/command/replSetGetConfig/#dbcmd.replSetGetConfig). This server command as I understand it can be run on any replica set node.

The procedure continues with rs.reconfig which links to [its server command documentation](https://docs.mongodb.com/manual/reference/command/replSetReconfig/#dbcmd.replSetReconfig). This command states the requirement to run it on the primary:

> You must run this command on the admin database of the primary replica set member.

The language suggested in this PR is actually incorrect - reconfiguration must be run on the primary, not on "any replica set member".

PR #2475, opened 2015-12-11
DOCS-6736: fix unit typography

ksuarz:
Ensures that a space is placed between a number and a unit; for example,
"64 GB".

In the GridFS documentation, "k" has been converted to "KB" when discussing
chunk size.

A hyphen is used for "32-bit" and "64-bit."


kay-kim:
merged

PR #2476, opened 2015-12-14
Update indexes-introduction.txt

divakarpatil51:
As per my understanding indexes exists for "Documents" and not for queries


kay-kim:
Hi -- thank you so much for your pull request.  I think the gist of the original sentence was that if an index exists that can support the query.  We can make that more explicit to avoid the confusion.  As such, we'll close this pull request, but much appreciate your efforts.

Regards,

Kay Kim

PR #2478, opened 2015-12-14
DOCS-6802: Add initial inMemory documentation

GeertBosch:


kay-kim:
Thanks Geert for the inmemory page as well as letting us know which options are no longer returned in the help.  I went ahead and removed those options.

PR #2479, opened 2015-12-14
DOCS-6736: reword atomicity and unique indexes

ksuarz:
This reworks the advice on utilizing unique indexes to achieve atomicity for certain operations.


kay-kim:
merged!

PR #2480, opened 2015-12-14
DOCS-6785: readConcern special cases

ksuarz:
Documents the special cases when using a read concern with a count or
aggregation command.


kay-kim:
merged

PR #2484, opened 2015-12-16
Fix a typo in the journalLatencyTest command name

kkmongo:
The command name does not have a leading underscore:

src/mongo/db/storage/mmap_v1/journal_latency_test_cmd.cpp:    JournalLatencyTestCmd() : Command("journalLatencyTest") {}


kay-kim:
Thanks Kamran!

PR #2488, opened 2015-12-22
DOCS-5204: clarify GridFS behavior for small files

ksuarz:
Clarifies the behavior of GridFS when storing files that are smaller than the
default chunk size.


kay-kim:
merged!

PR #2491, opened 2016-01-01
Update fact-3.0-downgrade-path.rst

Boomtime:
DOCS-6377 final agreed change
HELP-1450 for discussion
SERVER-17087 for the danger


kay-kim:
merged.  thank you.

PR #2492, opened 2016-01-03
Remove 's' regex flag from extended JSON

devkev:
I tested the mongo shell of every stable version between 1.6.0 and 3.2.0, and none support 's' as a valid flag.

---
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/2492)

### DIFF
--- a/source/reference/mongodb-extended-json.txt
+++ b/source/reference/mongodb-extended-json.txt
@@ -206,10 +206,9 @@ Regular Expression
   by the letters of the alphabet.
 
 - ``<jOptions>`` is a string that may contain only the characters 'g',
-  'i', 'm' and 's' (added in v1.9). Because the ``JavaScript`` and
-  ``mongo Shell`` representations support a limited range of options,
-  any nonconforming options will be dropped when converting to this
-  representation.
+  'i' and 'm'. Because the ``JavaScript`` and ``mongo Shell``
+  representations support a limited range of options, any nonconforming
+  options will be dropped when converting to this representation.
 
 OID
 ~~~


kay-kim:
Hey @devkev --
I'm able to use the s flag.
`~/Mongo $ ./mongodb-3.2.4/bin/mongoexport  -d test -c products --query '{ description: /m.*line/si }'`
`2016-05-02T10:33:59.813-0400	connected to: localhost`
`{"_id":102.0,"sku":"xyz456","description":"Many spaces before     line"}`
`{"_id":103.0,"sku":"xyz789","description":"Multiple\nline description"}`
`2016-05-02T10:33:59.813-0400	exported 2 records`
`~/Mongo $ ./mongodb-3.2.4/bin/mongoexport  -d test -c products --query '{ description: /m.*line/i }'`
`2016-05-02T10:34:04.620-0400	connected to: localhost`
`{"_id":102.0,"sku":"xyz456","description":"Many spaces before     line"}`
`2016-05-02T10:34:04.620-0400	exported 1 record`

devkev:
Ugh.  This had made me realise that this is actually a much bigger problem.

The issue is that mongo server/shell uses SpiderMonkey (in 3.2+, earlier versions use V8), whereas the tools use Golang - and each of these has their own RegExp implementations, with slight variations in the accepted flags.  eg, in addition to the normal `g`, `i`, `m`:
* SpiderMonkey has `u`, `y`:
  * https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/RegExp#Parameters
* Golang uses RE2, which has `s`, `U` (but not `g`)
  * https://golang.org/pkg/regexp/#pkg-overview
  * https://github.com/google/re2/wiki/Syntax

So this now feels like it deserves a DOCS ticket, rather than a "drive-by pull request".

kay-kim:
Hey Kevin -- 
will close this.  We do have a docs ticket  https://jira.mongodb.org/browse/DOCS-11478  to update the page.  Will see how to handle aforementioned issue when doing the ticket.

PR #2493, opened 2016-01-03
Change Extended JSON "OID" heading to "ObjectId"

devkev:
Since otherwise the only way that "OID" can be determined to actually mean "ObjectId", is by finding the mention buried in the Mongo shell mode sub-section.


kay-kim:
merged!  thank you!

PR #2497, opened 2016-01-06
[On Hold] DOCS-5324: quorum reads with findAndModify()

ksuarz:
This re-adds the procedure for accomplishing a quorum read -- that is, a read of data that is both committed and not stale -- through the use of the `findAndModify()` command.

This change is valid *only* for MongoDB 3.2 and later.


ksuarz:
Closing this pull request in favor of a cleaned-up, rebased version in a separate one.

PR #2498, opened 2016-01-07
Correct `updateMany` to `updateOne`

alexhouse:
Reference docs for `db.collection.updateOne` reference `updateMany` in the example. This corrects that.


schmalliso:
Merged! Thanks so much for the catch - we really appreciate it!

PR #2499, opened 2016-01-07
Update indexes.txt

kstirman:
Simple copy change, old copy had a typo.


schmalliso:
Merged. Thanks for the fix!

PR #2505, opened 2016-01-11
DOCS-6910: fixes grammar on agg pipeline intro paragraph

schmalliso:
The tiniest pull request ever -- we refer to 'aggregated results' elsewhere in the manual, but I wasn't sure if 'an aggregated result set' would be preferable.


kay-kim:
merged!

PR #2506, opened 2016-01-11
DOCS-6858: merge GridFS pages

ksuarz:
Merges /core/gridfs.txt and /reference/gridfs.txt to create one definitive
source of information on all things GridFS.


kay-kim:
merged!

PR #2510, opened 2016-01-11
DOCS-6872: reorganize storage and storage engines

ksuarz:
Reworded the storage section to give a more general overview of the different
storage sections in the Manual.

Moved the storage engines to their own sub-section underneath storage, complete
with short descriptions of each.


kay-kim:
merged

PR #2513, opened 2016-01-14
DOCS-6971: findAndModify + write concern error

ksuarz:
The comparison between `findAndModify` with the `update` command erroneously stated
that the former does not take a write concern. A write concern can be specified
starting in MongoDB 3.2.


kay-kim:
merged

PR #2514, opened 2016-01-14
DOCS-6879 in-memory storage engine and journaling

ksuarz:
Mentions the in-memory storage engine on the journaling page. (The behavior of
the in-memory storage engine with `{ j: 1 }` writes was already present on the
storage engine's page itself.)


kay-kim:
merged

PR #2515, opened 2016-01-14
DOCS-6922  improve enable-authentication.txt

rkumar-mongo:
User noted lack of direction in applying tutorial to a replica set or
sharded cluster. Added clarification that tutorial assumes a
standalone environment and linked to the replica set / sharded cluster
authentication tutorial.


kay-kim:
merged

PR #2517, opened 2016-01-14
DOCS-6874, DOCS-6942: read concern changes

ksuarz:
This pull request has a variety of read-concern fixes:

- Adds missing options for `db.collection.count()`, along with read concern
- Fixes the syntax of using a read concern with `count`
- Standardizes on "read concern" in writing and lines up the formatting with write concern


kay-kim:
merged

PR #2518, opened 2016-01-14
DOCS-5324: quorum reads with findAndModify()

ksuarz:
Details the process of performing *quorum* or *linearlizable* reads via the `findAndModify` command. The data read is guaranteed to be both durable and fresh if directed at the primary of a replica set.


kay-kim:
merged!

PR #2522, opened 2016-01-15
DOCS-6794 : Aggregation Reorg

rkumar-mongo:
R2: Refinements to Reference

R1:  Draft

@kay-kim there was a merge conflict in reference/operator/aggregation-pipeline.txt , I think I patched it but if you could double check me that would be great.


kay-kim:
merged!

PR #2525, opened 2016-01-18
Update glossary.txt

dryabtse:
Swapped words for the "write conflict" definition as it could originally be read as "MongoDB [server] will abort". Now it should be clear that only one of the conflicting operations will be aborted and retried.


kay-kim:
merged!

PR #2529, opened 2016-01-22
DOCS-6982: clarify result of upsert

ksuarz:
Explicitly say that an upsert will create a new document if it doesn't already
exist.


kay-kim:
merged!

PR #2531, opened 2016-01-22
Remove most of our intersphinx files

i80and:
Docs build is clean. I think we could also remove python2, but there's some brittleness in giza that requires us to have at least two entries here.


kay-kim:
merged!

PR #2533, opened 2016-01-26
Remove a link to the nonexistant geoNear query expression.

i80and:
Document validation only accepts query expressions, and the only prohibited
geospatial query expressions are near and nearSphere.


kay-kim:
merged!

PR #2535, opened 2016-01-27
DOCS-7000: no _id index scan for cursor.snapshot()

ksuarz:
Due to SERVER-19593, the `cursor.snapshot()` method no longer forces an `_id` index
scan and instead performs a collection scan.

This removes the description of the implementation of `cursor.snapshot()`. It also
removes the extract `isolate-cursor-snapshot-faq`.


kay-kim:
merged!

PR #2536, opened 2016-02-01
DOCS-5230: adds more detail on scheduling balancer across geographies

schmalliso:


source/includes/steps-schedule-balancer-window.yaml
```diff
@@ -0,0 +1,75 @@
...
+  - pre: |
+      The balancer will not activate in the ``stopped`` state.
+      To ensure that the  balancer
+      is not ``stopped``, use :method:`sh.setBalancerState()`,
+      as in the following:
+    language: sh
+    code: |
+      sh.setBalancerState( true )
+    post: |
+      The balancer will not start if you are outside
```
schmalliso: I did contemplate making this a `note` and adding more detail to why that's relevant (if you're updating the window, and you try to start it but are outside of the time, it won't start). Alternatively, we could maybe kill the `post` entirely.

kay-kim:
merged

PR #2537, opened 2016-02-01
Fix document validation example

jcarlosgarcia:
Fix a typo in the document validation example, using validationAction equals to "warn"


kay-kim:
Thank you so much.  You are indeed correct.  This has been merged. Regards.

PR #2544, opened 2016-02-05
Do not recommend logical config server DNS names for CSRS

devkev:
The idea of having logical DNS names for the config servers only applies the old-style triplet of config servers.  The rationale for it is being able to redirect the logical DNS name mapping to another host, in order to "swap out" one of the 3 config servers.  However, in the context of config servers *as replsets*, this is bad advice - any situation where you would "swap out" an old-style config server, with CSRS you should instead just use normal replica set management operations to handle (eg. add a new member, remove another member, etc).


kay-kim:
Thank you.  this has been merged.

PR #2557, opened 2016-02-26
DOCS-7216, 7271 - Fix images for write operations

rkumar-mongo:
* write-operations-introduction had incorrect text. Changed
'enrolled' to 'reject' for updateOne/updateMany and fixed case to be consistent with examples


kay-kim:
While we're at it, there is an additional error in the updateOne  (the sql condition and the mongodb condition do not match -- one shows greater than and the other shows less than )

Also, an error with the replaceOne -- the sql name and age fields do not match that of the mongodb example

rkumar-mongo:
It looks like the updateOne image was fixed in my initial commit - maybe at some point we uploaded a fixed but older version of updateOne.svg manually. 

Added replaceOne.svg with fixed text.

kay-kim:
merged

PR #2562, opened 2016-03-08
Clarify that you must flushRouterConfig after movePrimary for reads as well as writes

stbrody:


kay-kim:
merged! thanks Spencer

PR #2576, opened 2016-03-28
Use "gpgkey" in YUM repository configuration

kou:
It reduces an user action that running "rpm --import".


kay-kim:
Hi -- Thank you very much for this pull request.  This has been merged.  Much appreciated.

PR #2580, opened 2016-03-30
DOCS-7488: updated local key management section with correct log mess…

steveren:
updated local key management section with correct log message for version 3.2.4


kay-kim:
merged!

PR #2582, opened 2016-03-31
DOCS-7528: comment on "manual/reference/replica-states.txt"

steveren:
I wasn't sure whether to revert to the earlier commit or make a new commit, so I did the latter. It amounts to the same thing anyway...


kay-kim:
Oh, heh heh -- sorry, I use words in an ambiguous way.  You could have just git reset --hard HEAD~~  so that only your first commit stays.   But I can do that during my merge so no worries.  Thanks.

kay-kim:
merged.

PR #2592, opened 2016-04-06
Fixed creating neighborhoods  index

sallyx:
The neighborhoods document structure looks like {"_id": ..., "geometry" : { "coordinates": [ ...],"type" : "Polygon"},"name":"..."}}

It makes more sense to create index like db.neighborhoods.createIndex({ geometry: "2dsphere" }) than db.neighborhoods.createIndex({ coordinates: "2dsphere" })


kay-kim:
merged.  thank you!

PR #2603, opened 2016-04-18
DOCS-7678 recommend kernel.pid_max of 64000

devkev:
This is to match the ulimits nproc recommendation of 64000.

### DIFF
--- a/source/administration/production-checklist-operations.txt
+++ b/source/administration/production-checklist-operations.txt
@@ -192,7 +192,7 @@ Linux
    - Configure sufficient file handles (``fs.file-max``), kernel pid limit
      (``kernel.pid_max``), and maximum threads per process
      (``kernel.threads-max``) for your deployment. For large systems,
-     values of 98000, 32768, and 64000 are a good starting point.
+     values of 98000, 64000, and 64000 are a good starting point.
 
    - Ensure that your system has swap space configured. Refer to your
      operating system's documentation for details on appropriate sizing.


source/administration/production-checklist-operations.txt
```diff
@@ -192,7 +192,7 @@ Linux
    - Configure sufficient file handles (``fs.file-max``), kernel pid limit
      (``kernel.pid_max``), and maximum threads per process
      (``kernel.threads-max``) for your deployment. For large systems,
-     values of 98000, 32768, and 64000 are a good starting point.
+     values of 98000, 64000, and 64000 are a good starting point.
```
kay-kim: Hey @devkev  -- I just need to get rid of one of the two 64000 values, yes?

devkev: Uhhh, no?

The idea is that if pid_max is only 32768, then setting the nprocs ulimit to 64000 won't have the desired effect.  To put it another way, the user expectation is that after setting nprocs to 64000, they can have up to 64000 processes/threads on the system.  But this is only possible if pid_max is also >= nprocs, i.e. is >= 64000.  If it is less --- like the 32768 that we recommend here --- then the number of possible processes/threads will cap out less than 64000, which is confusing.

Therefore, we should ensure that we recommend setting pid_max to at least our recommended value for the nprocs ulimit.  Since we recommend nprocs be 64000, we should also recommend pid_max of 64000 --- which is what this patch does.

This is all assuming that the three values listed here correspond to file-max, pid_max and threads-max, respectively.  Feel free to rework things further to make that clearer, if necessary.

kay-kim: Ah, I see now.  I didn't realize that the three values listed corresponded to the  file-max, pid_max and threads-max : (

devkev: I think that's a vote for "it should be clearer" :)

kay-kim: indeed : )

kay-kim:
merged and tweaked the sentence to associate the value with the setting.

devkev:
Awesome, that looks great, thanks @kay-kim!!

kay-kim:
@devkev -- given that you did the heavy lifting, glad to do it : )

PR #2607, opened 2016-04-19
replace the word "evocation"

bgrabar:


bgrabar:
@thegreatco has suggested replacing the word "evocation," and after looking at the Wikipedia entry, I concur.

thegreatco:
We use "invocation" in many other places, I assume that was the intended word.

kay-kim:
merged

PR #2608, opened 2016-04-20
DOCS-7489: modern drivers use dbcommands instead of opcodes for ackno…

rkumar-mongo:
…wledged writes

### DIFF
--- a/source/reference/mongodb-wire-protocol.txt
+++ b/source/reference/mongodb-wire-protocol.txt
@@ -117,6 +117,14 @@ follows:
 Request Opcodes
 ~~~~~~~~~~~~~~~
 
+.. note:: 
+   Starting with MongoDB 2.6 and :data:`~isMaster.maxWireVersion` ``3``, 
+   drivers use the :ref:`database commands<collection-commands>` 
+   :dbcommand:`insert`, :dbcommand:`update`, and :dbcommand:`delete` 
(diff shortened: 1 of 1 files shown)


source/reference/mongodb-wire-protocol.txt
```diff
@@ -117,6 +117,14 @@ follows:
 Request Opcodes
 ~~~~~~~~~~~~~~~
 
+.. note:: 
+   Starting with MongoDB 2.6 and :data:`~isMaster.maxWireVersion` ``3``, 
+   modern drivers use the :ref:`database commands<collection-commands>`
```
kay-kim: by modern drivers, what do we mean?

rkumar-mongo: I would assume drivers that support MongoDB 2.6+. I think maxWireVersion should generally be at least MongoDB 3, right? So this might be better written as 'drivers supporting MongoDB 2.6+'

@ajdavis , does that sound about right?

ajdavis: Yeah, good point. I'd simply delete "modern".

rkumar-mongo: Done. I perused the drivers via ecosystem and none of the driver versions we list lack support for MongoDB 2.6. I'd assume by 'modern' it simply meant anything made within the last few years.

kay-kim:
merged

PR #2610, opened 2016-04-21
 DOCS-7476: Comment on: "manual/reference/method/db.cloneCollection.txt"

steveren:


source/reference/method/db.cloneCollection.txt
```diff
@@ -26,3 +26,8 @@ Behavior
 --------
 
 .. include:: /includes/extracts/cloneCollection-behavior-method.rst
+
+Example
+-------
```
kay-kim: Sorry -- just saw this single sourcing.  We're showing the command example instead of the method example here.  I don't think we can single source this.

source/includes/example-clone-collection.rst
```diff
@@ -0,0 +1,9 @@
+.. code-block:: javascript
+
+   { cloneCollection: "users.profiles",
```
kay-kim: We should make this a functioning example by surrounding it wiht

db.runCommand( { ... } )

kay-kim:
merged!

PR #2620, opened 2016-05-02
DOCS-5665, DOCS-5782, DOCS-6898, DOCS-7445: platform-specific considerations

schmalliso:
- clarify SELinux recommendation / that server-side JS does not work on SELinux
- recommend premium storage for Microsoft Azure
- generally tidy up the language and flow of the prodnotes' platform-specific
  considerations


kay-kim:
merged

PR #2633, opened 2016-05-19
DOCS-3854: emphasize uniform clusterAuthMode setting

rkumar-mongo:
* Also added callout noting internal auth enables
access control with callouts to the user tutorials


kay-kim:
merged!

PR #2638, opened 2016-06-03
Small correction about findOne uses.

davidrissato:
The text was saying findOne is used to remove documents instead of reading them.

---
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/2638)


schmalliso:
@davidrissato That is quite the glaring error! Thanks so much for the patch - we really appreciate it!

Merged 😸

PR #2644, opened 2016-06-09
DOCS-7979: Unclean shutdown may cause 60 seconds of drift in collStat…

rkumar-mongo:
…s or count

---
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/2644)


kay-kim:
merged

PR #2648, opened 2016-06-16
DOCS-7945: Fix the wrong index in the example

twang2218:
Based on the context, I think the first 2 `{ "user.login": 1 }` should be `{ "user": 1}`.

---
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/2648)

### DIFF
--- a/source/core/query-optimization.txt
+++ b/source/core/query-optimization.txt
@@ -159,9 +159,9 @@ An index **cannot** cover a query if:
 
   .. code-block:: none
 
-     { "user.login": 1 }
+     { "user": 1 }
 
-  The ``{ "user.login": 1 }`` index does **not** cover the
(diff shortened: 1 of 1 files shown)


kay-kim:
Hello -- 
thank you so much for taking the time to update and make this pull request. However, I do believe the example is correct; that is, the `{"user.login":1}` index supports the query but does not cover the query.  You can verify by doing an `explain()` on the find --  `db.users.find( { "user.login": "tester" }, { "user.login": 1, _id: 0 }).explain()`  -- the `explain` output includes an `IXSCAN` and a `FETCH` scan and a `PROJECTION` stage, indicating that the operation first used the index to find the matching documents, then fetched the matching documents, and finally did the projection.    As such, I will go ahead and close this pull request, but thanks again.  Much appreciated.

Regards,

Kay Kim

PR #2658, opened 2016-06-28
Mention that removeShard draining uses the balancer

devkev:
Follow-up from DOCS-7777.  That the balancer is the mechanism by which shards are drained is mentioned briefly in the Example.  This change makes the balancer's involvement much more explicit.

---
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/2658)


kay-kim:
merged!

PR #2662, opened 2016-07-07
DOCS-7935 Added upsert caveat that if querying on a specific value fo…

jdestefano-mongo:
…r _id without update operators and the upsert occurs, the _id will be added as part of the new document. -JD

DOCS-7935 Changed language for db.collection.update.  Added that update() and findAndModify() handle upserts similarly.  Updated parameter files to reflect upsert findings.  Removed incorrect statements for findOneAndReplace(). - JD

DOCS-7935 Fixed incorrectly named parameter.

---
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/2662)


kay-kim:
merged!

PR #2665, opened 2016-07-13
Username/password containing @ and : must be URL encoded

Boomtime:
Reference: https://github.com/mongodb/specifications/blob/master/source/connection-string/connection-string-spec.rst#userinfo-optional

Usernames and passwords that contain @ or : require these symbols to be URL encoded.

https://jira.mongodb.org/browse/DOCS-8333

---

This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/2665)

### DIFF
--- a/source/reference/connection-string.txt
+++ b/source/reference/connection-string.txt
@@ -51,6 +51,8 @@ The components of this string are:
        specific database using these credentials after connecting to
        the :program:`mongod` instance.
 
+       If username or password contains an at-sign ("@") or colon (":") then the username and/or password must be URL encoded. At-sign ("@") is replaced with "%40" (3 characters), while colon (":") is replaced with "%3A" (3 characters).
+
    * - ``host1``
 
      - Required. It identifies a server


source/reference/connection-string.txt
```diff
@@ -51,6 +51,8 @@ The components of this string are:
        specific database using these credentials after connecting to
        the :program:`mongod` instance.
 
+       If username or password contains an at-sign ("@") or colon (":") then the username and/or password must be URL encoded. At-sign ("@") is replaced with "%40" (3 characters), while colon (":") is replaced with "%3A" (3 characters).
+
```
kay-kim: Do you mind if we go more generic and then more specific example?

If the username or password contains reserved URI characters, you must escape the characters per RFC 2396. For example, if your username is s@m, you must escape the @ character when specifying the username in the connection string; i.e. s%40m.

kay-kim: Oh, I see that in an old driver spec file (https://github.com/10gen/specifications/commit/e22cd36938b184f3f98b0780b08ae4321becaac3) , they only talk about "@" and ":".  I'll go ahead and merge this in then.

kay-kim: Sorry for the noise, but it does seem to be more than just the "@" and ":"
I'll go with my first suggestion then.

kay-kim: I opened https://jira.mongodb.org/browse/SPEC-618  for the drivers + https://jira.mongodb.org/browse/NODE-757 since node.js didn't allow escaped colon in the username. I'll think of something for the docs tomorrow.  Thanks much.

Boomtime: Yes, I agree, it looks like all the drivers should be tested for this condition and bug reports raised for anything outside the spec. I think for the moment we use our specification as the definitive guide to what the common behavior should be (i.e escape @ and : symbols).

It would be nice to have the spec define what should happen if other escape sequences are present since people using Node (in particular, but also other languages) are likely to use standard libs to apply HTTP escaping, which will escape all manner of characters, not just 'at' and colon.

kay-kim:
Hey Andrew -- sorry for the delay, but we did finally get to DOCS-8333 so will close the pull.

PR #2669, opened 2016-07-18
DOCS-8337 - Fixed 'Update Document with Upsert' example by adding ups...

jdestefano-mongo:
…ert param to method call and ObjectId to result. Performed some page cleanup, changing mentions of 'replacement' to 'update'.

---
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/2669)


kay-kim:
merged

PR #2691, opened 2016-08-04
Fixed mongod systemd service path on Ubuntu 16.04

xet7:
Fixed mongod systemd service path on Ubuntu 16.04.

---
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/2691)


i80and:
This PR is absolutely correct, in that local administrator changes should go into /etc. Very good catch, and we appreciate it!

However, soon the packages will start including this file under /lib, and we don't want the temporary hack to override the official package systemd service files.

PR #2704, opened 2016-08-12
change "delete" to "update"

runME:
> "The following methods can also delete documents from a collection:"

     Not "delete",it should be "update" according to the context.

---
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/2704)


kay-kim:
merged!  Thank you!

PR #2730, opened 2016-09-27
DOCS-8276

rkumar-mongo:
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/2730)

### DIFF
--- a/source/administration/monitoring.txt
+++ b/source/administration/monitoring.txt
@@ -264,7 +264,7 @@ a paid subscription.
 
      - |MMS| is a cloud-based suite of services for managing MongoDB
        deployments. |MMS| provides monitoring, backup, and automation
-       functionality. For an on-premise solution, see also 
+       functionality. For an on-premise solution, see also
        :products:`Ops Manager, available in MongoDB Enterprise Advanced
        </mongodb-enterprise-advanced?jmp=docs>`.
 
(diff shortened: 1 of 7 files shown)


source/administration/monitoring.txt
```diff
@@ -354,6 +354,55 @@ affect logging:
...
 
+.. _monitoring-log-redaction:
+
+Log Redaction
+~~~~~~~~~~~~~
+
+.. versionadded:: 3.4 Available in MongoDB Enterprise only
+
+A :program:`mongod` running with :setting:`security.redactClientLogData`
+removes :doc:`messages </reference/log-messages>` associated with any given log
```
kay-kim: Q: does this remove messages?  as in it first logs and then redacts messages? or is it that the redacted message is logged.

source/includes/options-mongod.yaml
```diff
@@ -2164,4 +2164,37 @@ description: |
...
 optional: true
+---
+program: mongod
+name: redactClientLogData
+args: null
+directive: option
+description: |
+  .. versionadded:: 3.4 Available in MongoDB Enterprise only.
+  
+  A {{program}} running with {{role}} removes any message accompanying a given
```
kay-kim: same question with "remove" .

rkumar-mongo:
@kay-kim , pushed fixes plus an additional note. 

Note is based on continuing work to refine the output, so users aren't surprised if output changes at GA.

source/core/security-encryption-at-rest.txt
```diff
@@ -99,6 +99,25 @@ transport encryption.
 
 For details, see :ref:`rotate-encryption-keys`.
 
+Logging
+~~~~~~~
+
+.. versionadded:: 3.4 Available in MongoDB Enterprise only
```
kay-kim: Just wondering:  Given that under the Encrypted Storage Engine header we state that the feature is Enterprise only, do we need to reiterate the enterprise availability here as well?

kay-kim:
merged!

PR #2731, opened 2016-09-27
Reorder rs node removal instructions

milkie:
To avoid losing quorum, one should always run reconfigure prior to shutting down nodes marked for removal.

---
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/2731)


kay-kim:
Thanks mister!

PR #2737, opened 2016-10-11
Don't use 'force' when stepping down config servers for CSRS->SCCC do…

stbrody:
…wngrade

---
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/2737)


kay-kim:
merged. Thanks mister!

PR #2740, opened 2016-10-18
DOCS-9124 - Added a note in the BSON types table that decimal was added in version 3.4.

jdestefano-mongo:
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/2740)


kay-kim:
merged

PR #2743, opened 2016-10-20
Update unset.txt

smitthakkar96:
There is nowhere specified that if multi is default set to false and will only affect 1 document. It's a bit confusing.

---

This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/2743)

### DIFF
--- a/source/reference/operator/update/unset.txt
+++ b/source/reference/operator/update/unset.txt
@@ -21,7 +21,9 @@ $unset
 
    The specified value in the :update:`$unset` expression (i.e. ``""``)
    does not impact the operation.
-
+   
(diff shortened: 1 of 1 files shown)


kay-kim:
Hi --  thank you so much for this pull request and apologies for the lateness of this response.
Because the {multi: true} is an option that is set at the db.collection.update() method rather than individual operators available for the update, the option isn't mentioned on the individual pages.

There will be a future effort to revamp the various operator pages, and details such as this will be incorporated in the revamp.  As such, will close this pull request without merge as we'd want to tackle the project as a whole, but we very much appreciate your taking the time to submit a pull request.  

Regards,

Kay Kim

PR #2749, opened 2016-10-28
Fix the returnOriginal param in the findOneAndUpdate

ahmadassaf:
The correct parameter to return the new update document when using `findOneAndUpdate` is `returnOriginal` instead of `returnNewDocument`

---

This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/2749)


kay-kim:
Hi -- 
thank you for this pull request and apologies for the long delay.  
I believe the returnOriginal is for the node.js driver whereas the page in question is for the mongo shell method where returnNewDocument is correct (we have since updated the page to distinguish between the two).  As such, will close this pull request, but we definitely appreciate your taking the time to help us improve the docs and apologies once again for the delay in response.

Regards, 

Kay Kim

PR #2759, opened 2016-11-09
DOCS-7700

steveren:
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/2759)


kay-kim:
Hey Steve -- there's an errant commit DOCS-8822 in this pull request.  Are you following the workflow we went over as documented in the wiki?  No need to fix in this case.  I've fixed during merge but wanted to make sure.

PR #2766, opened 2016-11-14
DOCS-7589: Enable JavaScript Protection by default

steveren:
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/2766)

### DIFF
--- a/source/includes/options-mongo.yaml
+++ b/source/includes/options-mongo.yaml
@@ -350,4 +350,56 @@ description: |
    .. versionadded:: 3.2
 
    Disables use of the JavaScript engine's JIT compiler.
+---
+program: mongo
+name: disableJavaScriptProtection
+directive: option
+optional: true
+args: null
+description: |
+   .. versionadded:: 3.4
+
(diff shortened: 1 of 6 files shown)


source/includes/options-mongo.yaml
```diff
@@ -350,4 +350,57 @@ description: |
...
+directive: option
+optional: true
+args: null
+description: |
+   .. versionadded:: 3.4
+
+   Allows documents containing fields of type
+   :ref:`javascript<bson-types>` and
+   :ref:`javascriptWithScope<bson-types>` to be automatically
+   marshalled to JavaScript functions in the :program:`mongo`
```
kay-kim: Are we marshalling the whole documents to JavaScript functions?  Because that's what this sentence states.

steveren: No, it's only the fields of type javascript and javascriptWScope which are marshalled to the shell. I've pushed a new version of this branch with updated wording.

kay-kim:
merged!

PR #2778, opened 2016-11-22
DOCS-8290 - Added manual page for bucket pipeline stage.

jdestefano-mongo:
Minor edits to bucketAuto and sortByCount descriptions on ToC and manual pages.

---
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/2778)

### DIFF
--- a/source/includes/ref-toc-aggregation-pipeline-operator.yaml
+++ b/source/includes/ref-toc-aggregation-pipeline-operator.yaml
@@ -113,21 +113,26 @@ description: |
    aggregations capable of characterizing data across multiple 
    dimensions, or facets, in a single stage.
 ---
+name: :pipeline:`$bucket`
+file: /reference/operator/aggregation/bucket
+description: |
+   Categorizes incoming documents into groups, called buckets, based on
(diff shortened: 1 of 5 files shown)


source/includes/ref-toc-aggregation-pipeline-operator.yaml
```diff
@@ -113,21 +113,26 @@ description: |
    aggregations capable of characterizing data across multiple 
    dimensions, or facets, in a single stage.
 ---
+name: :pipeline:`$bucket`
+file: /reference/operator/aggregation/bucket
+description: |
+   Categorizes incoming documents into groups, called buckets, based on
+   a specified expression and bucket boundaries.
+---
```
jdestefano-mongo: Alternatives:

Categorizes incoming documents into groups, called buckets, based on a specified expression. Each bucket has a user-specified range of the expression's values.

Categorizes incoming documents into groups, called buckets, based on a specified expression and the boundaries of each bucket.


If one of these sounds better and is used instead, change the first sentence of bucket.txt as well.

kay-kim:
merged

PR #2783, opened 2016-11-25
Update glossary.txt

joannac-:
Initial sync is the process of copying all data to a new member. The "or restored" part is confusing - a member that has had data restored should not need to go through the initial sync process (unless the restore was not valid)

---
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/2783)


kay-kim:
merged!  Thank you.

PR #2787, opened 2016-11-28
DOCS-8262: relnotes for system certs

schmalliso:
I've not yet actually done this ticket, but it seemed useful to include it in the relnotes.

---
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/2787)


kay-kim:
merged

PR #2797, opened 2016-12-06
DOCS-7787: Document isMaster.lastWrite

i80and:
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/2797)

### DIFF
--- a/source/reference/command/replSetGetStatus.txt
+++ b/source/reference/command/replSetGetStatus.txt
@@ -78,7 +78,7 @@ Consider the following example output:
                 "t" : NumberLong(1)
              }
           },
-       
+
        "members" : [
            {
                "_id" : 0,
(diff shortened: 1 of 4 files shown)


kay-kim:
@i80and  -- Is there somewhere that this is staged?  I just wanted to see how the include file reads in its context.

i80and:
https://docs-mongodbcom-staging.corp.mongodb.com/andrew/DOCS-7787/reference/command/isMaster.html#dbcmd.isMaster

kay-kim:
Not sure if this flows quite right:
https://docs-mongodbcom-staging.corp.mongodb.com/andrew/DOCS-7787/reference/command/replSetGetStatus.html#replSetGetStatus.optimes

i80and:
Eesh, you're right; that file's change was an after-thought, and it shows.

Fixed, but let me know if you want me to do another round of review.

source/reference/glossary.txt
```diff
@@ -585,6 +585,18 @@ Glossary
       basic mechanism enabling :term:`replication` in MongoDB.
       See :doc:`/core/replica-set-oplog`.
 
+   optime
+      .. versionchanged:: 3.2
+
+         MongoDB 3.2 introduced replication
+         :rsconf:`protocolVersion: 1 <protocolVersion>` with a new optime
+         format.
```
i80and: I see. How about

    The following describes the optime format used by `protocolVersion: 1`, introduced by MongoDB 3.2.

i80and: We use "introduced in" in our existing corpus, and is generally a more common form I feel. Updated.

source/reference/command/replSetGetStatus.txt
```diff
@@ -195,23 +195,13 @@ following fields:
...
-   progress for various operations.
-   
-   If using :rsconf:`protocolVersion: 1 <protocolVersion>`, the optime
-   value is a document that contains:
-
-   - ``ts``, the :ref:`Timestamp <document-bson-type-timestamp>` of
-     the operation.
+   The ``optimes`` field holds a document that contains
+   :term:`optimes <optime>` showing replication progress for various
+   types of operations.
```
i80and: As far as I can see, there are three main differences:

- s/information on various optimes/optimes
- s/reported here to show/showing
- No longer notes different behavior with protocol version 0

The first is only incorrect if the following documentation is also wrong: the documented document fields are just optimes.

kay-kim: So, it's the second point:
optimes showing replication progress ...  makes it sound like there are optimes that show one thing and other optimes showing some other optime.

Also, separately, both the original wording and new version for "various operations" and "various types of operations" -- I wonder if we should just get rid of it as  it sounds like there are different optimes for updates or something.

kay-kim:
merged!

PR #2801, opened 2016-12-09
DOCS-6308: Elaborate on iptables w/ HTTP interface

i80and:
Also remove some weird content duplication and fix some grammar while I'm in here.

---
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/2801)


source/tutorial/configure-linux-iptables-firewall.txt
```diff
@@ -218,16 +209,16 @@ Provide Access For Monitoring Systems
 
    .. optional::
 
-      For config server :program:`mongod` instances running with
-      the ``shardsvr`` value for the :setting:`~sharding.clusterRole` 
+      For shard server :program:`mongod` instances running with
```
kay-kim: This could be read as shard server mongod instances running with shardsvr, and hence there might be shard server mongodb instances not running with shardsvr.

same for below.

kay-kim:
merged!

PR #2803, opened 2016-12-12
add arch tag to avoid apt update error/warning

amongouser:
The warning is:

N: Skipping acquire of configured file 'multiverse/binary-i386/Packages' as repository 'http://repo.mongodb.org/apt/ubuntu xenial/mongodb-org/3.4 InRelease' doesn't support architecture 'i386'

---
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/2803)


i80and:
Thanks for the PR!

This will break non-x86 platforms. I have a more general fix that should go out later this week.

PR #2815, opened 2017-01-10
Use "mongod processes" rather than "mongod servers"

lindaqin:
Since the word "server" can often be misinterpreted as the host itself, giving different meaning to "restart the mongod server".

---
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/2815)


kay-kim:
merged!  Thank you.

PR #2821, opened 2017-01-17
DOCS-9177: removes --upgrade option from mongos ref

schmalliso:
`mongos --upgrade` was removed in 3.2 (it's also absent in 3.4, unsurprisingly). 

Patch should be backported to 3.2 as well, please! 😺

---
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/2821)


kay-kim:
merged to 3.2 and master

PR #2823, opened 2017-01-18
DOCS-9744: $sample can only use a random cursor if it is the first st…

steveren:
…age in the pipeline

---
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/2823)


kay-kim:
merged

PR #2828, opened 2017-01-23
DOCS-9766-v3.0: backport validate() changes for v3.0 and v3.2

steveren:
Note: I tested validate() with 3.0 and 3.2 and found the same output as with 3.4, so I'm updating those versions of the manual with the same page that's in 3.4.

---
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/2828)


kay-kim:
merged

PR #2838, opened 2017-02-14
DOCS-9481: shell mkdir should return whether a new directory was created

steveren:
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/2838)

### DIFF
--- a/source/reference/method/mkdir.txt
+++ b/source/reference/method/mkdir.txt
@@ -17,10 +17,46 @@ Description
 
    Creates a directory at the specified path. This method creates
    the entire path specified if the enclosing directory or
-   directories do not already exit.
+   directories do not already exit. The user running the
+   :program:`mongo` shell must have permission to create directories in
+   the specified path.
 
    This method is equivalent to :command:`mkdir -p` with BSD or GNU utilities.
 
    The :method:`mkdir()` method has the following parameter:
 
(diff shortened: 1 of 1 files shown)


source/reference/method/mkdir.txt
```diff
@@ -17,10 +17,41 @@ Description
...
+
+Example
+-------
+
+The following command creates a directory called ``foo`` in the
+shell's current working directory.
+
+.. code-block:: javascript
+
+   > mkdir("foo")
```
kay-kim: Oh, we try not to put the ">"  in our code because once we make all code examples as copyable, it would copy the prompt as well.

steveren:
Took out the shell prompt.

source/reference/method/mkdir.txt
```diff
@@ -17,10 +17,41 @@ Description
...
+Example
+-------
+
+The following command creates a directory called ``foo`` in the
+shell's current working directory.
+
+.. code-block:: javascript
+
+   mkdir("foo")
+   { "exists" : true, "created" : true }
```
kay-kim: We should separate out code they should run and the results.  Otherwise, the copyable code-block would include the results which would make pasting weird.

steveren:
Separated command from results.

kay-kim:
mergeD!

PR #2840, opened 2017-02-16
add missing field name

asya999:
Example less confusion when showing that location-field is required.

---
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/2840)


kay-kim:
merged!

PR #2848, opened 2017-02-28
Typo corrected

bukkasamudram:
As per my understanding, If the application can tolerate non-atomic updates for two pieces of data, you can store these data in separate collections.

---
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/2848)


kay-kim:
Hi.  Thank you for this pull request and apologies for the long delay.  

While your statement is true, in context of the paragraph, I do believe that "separate documents" is appropriate (the separate documents can be in the same or different collections).  As such, I will close this pull request.  However, we greatly appreciate your time and effort to improve the docs.   And apologies once again for the lateness of the response.

Regards,

Kay Kim

PR #2849, opened 2017-03-02
$count aggregator returns array

miccarr:
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/2849)

### DIFF
--- a/source/reference/operator/aggregation/count.txt
+++ b/source/reference/operator/aggregation/count.txt
@@ -74,4 +74,4 @@ The operation returns the following results:
 
 .. code-block:: javascript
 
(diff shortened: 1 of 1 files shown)


kay-kim:
Hi.   Thank you for this pull request and apologies for the delay. We appreciate your taking the time to make this pull request.

However, I believe the $count example is correct and the aggregate method returns a cursor to the results (in the mongo shell, if the returned cursor is not assigned to a variable using the var keyword, then the cursor is automatically iterated up to 20 time).

As such, I am going to close this pull request. 

Thanks once again and apologies for the delay.

Regards,

Kay Kim

PR #2851, opened 2017-03-03
Typo corrected

venkicnu:
In case of maxStalenessSeconds mentioned as part read preference, the reads will be redirected to the secondary whose estimated lag less than the maxStalenessSeconds value and not greater than maxStalenessSeconds value.

---
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/2851)


venkicnu:
Please ignore this change.

PR #2860, opened 2017-03-30
Docs 8007

steveren:
There were a couple of broken substitutions as a result of this ticket, which NewAndrew spotted and pointed out. I altered the include file to remove the substitution directive and updated the files which use the include file.

---
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/2860)


kay-kim:
merged!

PR #2863, opened 2017-04-06
Removed ToC on Query for Null page and added Python None caveat for…

jdestefano-mongo:
…Equality Filter section

---
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/2863)


kay-kim:
merged!

PR #2864, opened 2017-04-07
minor: increase depth of server parameters TOC to 2

ksuarz:
The table of contents on the server parameters page isn't that useful, since it only points to the top-level headings. It would be nice if it also had quick links to jump to all of the subheadings.

I haven't built this locally or tested it in staging, so sorry if this is wrong :)

---
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/2864)


kay-kim:
merged!   Thank you!

PR #2869, opened 2017-04-12
DOCS-10032 New section for tag set write concern behavior

kevinadi:
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/2869)


kay-kim:
merged!  Thanks much!

PR #2877, opened 2017-05-05
DOCSP-607: tabbed examples for Ruby

steveren:
I assume the 'can't automatically merge' is the result of multiple updates to conf.py and Makefile and such. If there's a better way to do this, just let me know.

---
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/2877)


kay-kim:
No -- that's fine.  You did accidentally overwrite the perl tab in example #53 but I'll put it back in another commit.

PR #2881, opened 2017-05-16
DOCS-10212 provided an example when rs.add will cause an election

kevinadi:
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/2881)


kevinadi:
merged

PR #2882, opened 2017-05-16
Add note about extended localhost exception in 3.4

kirbyk:
CR: @i80and

---
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/2882)

### DIFF
--- a/source/core/security-users.txt
+++ b/source/core/security-users.txt
@@ -145,6 +145,14 @@ the first user in the ``admin`` database. The first user must have
 privileges to create other users, such as a user with the
 :authrole:`userAdmin` or :authrole:`userAdminAnyDatabase` role.
 
+.. versionchanged:: 3.4
+
+   The localhost exception was extended to include the execution of the
+   :method:`db.createRole()` method. This provides users authorizing via LDAP
+   the ablility to create a role inside of MongoDB that maps to a role defined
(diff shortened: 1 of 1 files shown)


source/core/security-users.txt
```diff
@@ -145,6 +145,14 @@ the first user in the ``admin`` database. The first user must have
 privileges to create other users, such as a user with the
 :authrole:`userAdmin` or :authrole:`userAdminAnyDatabase` role.
 
+.. versionchanged:: 3.4
+
+   The localhost exception was extended to include the execution of the
```
i80and: We try to avoid passive voice when there's a straightforward active voice form. In this case, you would probably want `MongoDB 3.4 extended the localhost exception to...`

source/core/security-users.txt
```diff
@@ -145,6 +145,14 @@ the first user in the ``admin`` database. The first user must have
...
+   :method:`db.createRole()` method. This provides users authorizing via LDAP
```
i80and: This phrasing could be slightly tighter. Consider `This method allows users authorizing via LDAP to create...`

i80and:
Merged in f393cf76e4b627e59e760774ee0f69358c72e7cf

Thanks, ᴋɪʀʙʏ :D

PR #2885, opened 2017-05-25
DOCS-8915: 'as' option of $filter and $map defaults to 'this'

nlarew:
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/2885)

### DIFF
--- a/source/reference/operator/aggregation/filter.txt
+++ b/source/reference/operator/aggregation/filter.txt
@@ -17,7 +17,7 @@ Definition
 
    .. versionadded:: 3.2
 
-   Selects a subset of the array to return based on the specified
+   Selects a subset of an array to return based on the specified
    condition. Returns an array with only those elements that match the
    condition. The returned elements are in the original order.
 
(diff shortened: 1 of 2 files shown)


source/reference/operator/aggregation/filter.txt
```diff
@@ -41,10 +41,10 @@ Definition
...
       * - ``as``
 
-        - The variable name for the element in the ``input`` array. The
-          ``as`` expression accesses each element in the ``input``
-          array by this :doc:`variable
-          </reference/aggregation-variables>`.
+        - Optional. The variable name for the element in the ``input`` array. The
+          ``cond`` expression accesses each element in the ``input`` array
+          by this :doc:`variable </reference/aggregation-variables>`.
+          If not specified, ``as`` defaults to ``$$this``.
```
kay-kim: Not sure, but should it be "this"? since in the cond expression, we'll access the variable as $$this, so the value of "as" would be "this".

nlarew: Yes I think "this" is a bit clearer. Updated in 1a310637f55e395446de7636072afc7ffb74eb5e

source/reference/operator/aggregation/filter.txt
```diff
@@ -41,16 +41,17 @@ Definition
...
 
       * - ``cond``
 
-        - The :ref:`expression <aggregation-expressions>` that
-          determines whether to include the element in the resulting
-          array. The expression accesses the element by the variable
+        - An :ref:`expression <aggregation-expressions>` that resolves
+          to a boolean value used to determine if an element should be
+          included in the output array. The expression references each
+          element of the ``input`` array inidividually with the variable
```
kay-kim: minor typo -- inidividually.

nlarew:
Pushed c3f899ba9ac19f298029bd5c2b8f5ae28c3cab0d with a fix for the typo.

kay-kim:
merged!

PR #2900, opened 2017-06-20
Use the default port for config servers

Wilfred:
By default, a config server listens on port 27019, so use that in the example.

---
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/2900)


kay-kim:
Hi. Thank you so much for this catch.  It seems like we reference the config server with 27017 in other places on that page (when we specify the configservers in the mongos).  As such, in order to do a clean sweep, I've opened https://jira.mongodb.org/browse/DOCS-10417 to do all the changes at once. 

As such, I will close this pull without merge.  But thank you once again for catching this and taking the time to fix and make this pull request.

Warmest regards,

Kay Kim

PR #2910, opened 2017-06-28
Docs 10365 - Minor text fix for string length example

jeff-allen-mongo:
Correcting the string length of 'cafeteria' to be 9 instead of 11.

CR: https://mongodbcr.appspot.com/141360001/

---
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/2910)


kay-kim:
merged!

PR #2911, opened 2017-06-28
DOCS-10366 Adding missing field to doc example

jeff-allen-mongo:
Minor text change - 'year' field has been added to the example document.

CR: https://mongodbcr.appspot.com/137640001/

---
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/2911)


jeff-allen-mongo:
Hey Kay - Sorry about that, I misunderstood what the ticket was asking for, should be all set now. 

A thought I had was that in the example documents _id: 7 explicitly states that there is no price via an inline comment.
Do we want to also explicitly state that this doc (_id: 4) has no year in a similar fashion?

kay-kim:
merged!

PR #2929, opened 2017-08-03
Update db.collection.validate.txt

joannac-:
"full" is the name of the parameter, but the value that should be passed in is "true"

---
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/2929)


kay-kim:
merged!  Thank you

PR #2933, opened 2017-08-10
Grammar fixes in ObjectId description

jmikola:
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/2933)

### DIFF
--- a/source/reference/bson-types.txt
+++ b/source/reference/bson-types.txt
@@ -37,8 +37,8 @@ ObjectId
 --------
 
 ObjectIds are small, likely unique, fast to generate, and ordered.
-ObjectId values consists of 12-bytes, where the first four bytes are a
-timestamp that reflect the ObjectId's creation, specifically:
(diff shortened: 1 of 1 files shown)


source/reference/bson-types.txt
```diff
@@ -37,8 +37,8 @@ ObjectId
 --------
 
 ObjectIds are small, likely unique, fast to generate, and ordered.
-ObjectId values consists of 12-bytes, where the first four bytes are a
-timestamp that reflect the ObjectId's creation, specifically:
+ObjectId values consist of 12-bytes, where the first four bytes are a
```
jmikola: @kay-kim: I'm not sure if "12-bytes" is correct here. In context, "12 bytes" may be more appropriate, as we're not using "12-bytes" as an adjective.

kay-kim: You are correct sir!  12 bytes.  Made an update after merging your pull.  Thank you!

kay-kim:
merged!

PR #2939, opened 2017-08-22
DOCS-9528: mmapv1 does not run on big endian architectures

schmalliso:
Needs back-porting to 3.4 as well.

---
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/2939)


kay-kim:
merged to both

PR #2960, opened 2017-09-20
DOCS-10113 collMod and planCacheSetFilter use index name

pvselvan:
if multiple indexes have same key pattern, you must specify an index name instead of the specification document

---
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/2960)


kay-kim:
merged!

PR #2961, opened 2017-09-20
DOCS-10525: Timezones in date aggregation operators

nlarew:
[**Jira Ticket**](https://jira.mongodb.org/browse/DOCS-10525)
[**Staged Changes**](https://docs-mongodbcom-staging.corp.mongodb.com/nick/DOCS-10525/reference/operator/aggregation-date.html)
[**Code Review**](https://mongodbcr.appspot.com/160310001/)

Adds documentation for timezone conversion on date aggregation operators.
All new behavior examples have been tested on v3.5.12 and are confirmed to be accurate.

NOTE: This only covers existing timezone operators. The new date operators for 3.6 are being documented in [DOCS-10475](https://jira.mongodb.org/browse/DOCS-10475) and may require a quick pass for consistency with these changes.

---
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/2961)

PR #2964, opened 2017-09-25
Remove GMT from tzExpression default

nlarew:
Per @atsansone's comment on https://github.com/mongodb/docs/pull/2961

> Note: UTC is a time standard and GMT is a UK time zone. When daylight savings time is in effect in the UK, they go to BST (British Standard Time). You should refer to this as UTC only. (https://www.timeanddate.com/time/gmt-utc-time.html)

---
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/2964)


kay-kim:
thanks mister.  merged!

PR #2970, opened 2017-09-28
Add clarification about compound multikey indexes

edaniels:
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/2970)

### DIFF
--- a/source/core/index-multikey.txt
+++ b/source/core/index-multikey.txt
@@ -91,6 +91,31 @@ an array; i.e. no document contains array values for both ``a`` and
 attempt to insert a document where both ``a`` and ``b`` fields are
 arrays, MongoDB will fail the insert.
 
+Indexing multiple fields within the same array field is also permitted. Consider
+a collection that contains the following documents:
+
+.. code-block:: javascript
+
+   { _id: 1, a: [ { "sub_field_a": 1, "sub_field_b": 2 }, { "sub_field_b": 1 } ],  category: "A array" }
+   { _id: 1, a: [ { "sub_field_a": 1 }, { "sub_field_b": 2 } ],  category: "A array" }
+
+A compound multikey index ``{ "a.sub_field_a": 1, "a.sub_field_b": 1 }`` is
+permissible. A multikey index entry will be created for each element consisting
+of the two subfields. If the index were unique, the following document would fail
+to be inserted:
+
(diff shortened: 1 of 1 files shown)


edaniels:
@kay-kim my assertion about uniqueness is not true for `2dsphere`. Is it worth noting that?

kay-kim:
heh -- probably.

kay-kim:
Am tweaking the wording a bit.  I can stick the blurb about 2dsphere in if you want.

edaniels:
Sure. Cool with me

source/core/index-multikey.txt
```diff
@@ -91,6 +91,31 @@ an array; i.e. no document contains array values for both ``a`` and
...
+
+   { _id: 3, a: [ { "sub_field_b": 1 }  ],  category: "A array" }
+   
+However, the following document would be inserted because the multikey index is on
+each element, not the cartesian product of each field in the array referenced by the
+index:
+
+.. code-block:: javascript
+
+   { _id: 3, a: [ { "sub_field_a": 1 }, { "sub_field_b": 1 } ],  category: "A array" }
```
kay-kim: @edaniels  --  doublechecking this wouldn't be true.  Since the second document above (line 100) would already have an index entry { a.sub_field_a: 1, a.sub_field_b: null } as well as { a.sub_field_a: null, a.sub_field_b: 1 } 

Did you mean  [ { sub_field_a: 1, sub_field_b: 1 } ]  ?

edaniels: Sorry. Line 100 should change to an _id of 2. Doing that will make 1 and 2 insert. 3 will fail to insert with 
`E11000 duplicate key error collection: test.test index: a.sub_field_a_1_a.sub_field_b_1 dup key: { : null, : 1.0 }`

kay-kim: Yup -- but double checking how are the two _id:3 documents different?  They both would fail because of uniqueness constraint?

edaniels: Sorry Kay, you're right: This doc here should be
`{ _id: 3, a: [ { "sub_field_a": 1, "sub_field_b": 1 } ],  category: "A array" }`

kay-kim:
closing as it's merged into local branch.

PR #3021, opened 2017-10-16
DOCS-10737: server parameter honorSystemUmask

steveren:
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/3021)

### DIFF
--- a/source/reference/parameters.txt
+++ b/source/reference/parameters.txt
@@ -1251,6 +1251,32 @@ Storage Parameters
    .. seealso:: :setting:`~storage.syncPeriodSecs` and
       :parameter:`journalCommitInterval`.
 
+.. parameter:: honorSystemUmask
+
+   .. versionadded:: 3.6
+
+   *Default*: ``false``
+
+   If :parameter:`honorSystemUmask` is set to ``true``, new files
+   created by MongoDB have permissions in accordance with the
+   user's ``umask`` settings.
+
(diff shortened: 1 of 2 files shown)


steveren:
Staged: https://docs-mongodbcom-staging.corp.mongodb.com/steverenaker/DOCS-10737/release-notes/3.6.html#mongod-options

For the release notes, I created a new section for new mongod options (since there are many in 3.6) and used the same formatting as currently exists for the new Sessions parameters. If it would be better to make a table with the option name on the left and a description on the right, let me know and I'll make it so.

kay-kim:
Ah -- this would for the present just go under the general enhancements.
We tend to subsection the key features -- sessions/(we'll have change streams)/agg ...
But for new options for mongod, if the new options are specific to some feature, then we'll group those instead of a section on new mongod options since some new options for mongod could be available in mongos and mongo shell.  Close to when we have the bulk of things doc'ed, I think your sectioning probably is the way to go so that we list the options in couple of places on that page. But for now, could we just add a single bullet point under the general enhancements?  That way, we can control when we do the reorging (vs. someone opening a ticket now going 'Hey, you haven't listed some of the new options under that section').

steveren:
Thanks Kay. Updated as you suggested.

kay-kim:
merged!  Thanks Steve.

PR #3026, opened 2017-10-17
DOCS-9933: Regex is unable to take advantage of indexes that use collation

stennie:
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/3026)


kay-kim:
merged!

PR #3030, opened 2017-10-19
DOCSP-1008: 3.6: adds examples to positional-filtered and positional-all operators

schmalliso:
- Adds upsert example to positional-filtered
- Adds negation examples to both positional operators

---
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/3030)


kay-kim:
merged!

PR #3047, opened 2017-10-24
DOCS-10228 intra cluster compression is not on by default in 3.4.6

pvselvan:
Ticket didn't make it through code review while on Docs rotation, currently being reviewed by Jonathan Reams.

---
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/3047)


kay-kim:
merged into a local branch from which we'll continue.

PR #3056, opened 2017-10-26
DOCS-10935: hacky solution to make non-linking methods link

schmalliso:
For whatever reason, giza/sphinx isn't building links for a bunch of methods. This is a hacky solution to make the links display on the /reference/methods page (though they're not literals, it seems more practical to have a link and no-monospace than the alternative). It doesn't seem worth the effort to figure out why sphinx is behaving weirdly.

---
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/3056)


schmalliso:
Merged to master & v3.4. Ty @i80and!

PR #3060, opened 2017-10-28
(DOCS-10944, -10705): Added limitations of --oplog and --oplogReplay.

atsansone:
@rkumar-mongo : Could you give this a look?

---
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/3060)

### DIFF
--- a/source/includes/options-mongodump.yaml
+++ b/source/includes/options-mongodump.yaml
@@ -254,11 +254,13 @@ description: |
   {{role}} does not dump the oplog collection.
 
   .. note::
+
      :program:`mongodump --oplog <mongodump>` must create a full dump of a
-     ::term:`replica set` member. You cannot use :option:`--db` or
-     ::option:`--collection` with {{role}} to limit what will be dumped. If
-     :you use either of these options with {{role}}, the dump fails and
-     :generates an error message.
+     :term:`replica set` member. :program:`mongodump` fails if you use any of
+     the following options with {{role}} to limit the data to be dumped:
+
+     - :option:`--db`
+     - :option:`--collection`
 
   .. seealso:: :option:`mongorestore --oplogReplay <--oplogReplay>`
 
--- a/source/includes/options-mongodump.yaml
+++ b/source/includes/options-mongodump.yaml
@@ -255,9 +255,9 @@ description: |
 
   .. note::
 
-     :program:`mongodump --oplog <mongodump>` must create a full dump of a
-     :term:`replica set` member. :program:`mongodump` fails if you use any of
-     the following options with {{role}} to limit the data to be dumped:
+     To use :program:`mongodump` with {{role}}, you must create a full dump of
+     a :term:`replica set` member. :program:`mongodump` with  {{role}} fails
+     if you use any of the following options to limit the data to be dumped:
 
      - :option:`--db`
      - :option:`--collection`
--- a/source/includes/options-mongodump.yaml
+++ b/source/includes/options-mongodump.yaml
@@ -252,6 +252,16 @@ description: |
   :term:`master` nodes in master/slave replication deployments.
 
   {{role}} does not dump the oplog collection.
+
+  .. note::
+     :program:`mongodump --oplog <mongodump>` must create a full dump of a
+     ::term:`replica set` member. You cannot use :option:`--db` or
+     ::option:`--collection` with {{role}} to limit what will be dumped. If
+     :you use either of these options with {{role}}, the dump fails and
+     :generates an error message.
+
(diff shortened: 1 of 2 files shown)


source/includes/options-mongorestore.yaml
```diff
@@ -239,6 +240,11 @@ description: |
   level of the dump directory for :program:`mongorestore --oplogReplay
   <mongorestore>` to replay.
 
+  :program:`mongorestore --oplogReplay <mongorestore>` must restore the
```
rkumar-mongo: I think this is the other way around. ":program:`mongorestore --oplogReplay <mongorestore>` requires a full database dump of a :term:`replica set` member"

In fact, it might specifically require the `--oplog` flag to mongoexport. 

Then you can continue with ":program:`mongorestore --oplogReplay <mongorestore>` can only perform full database restorations. You cannot limit the restoration process using options such as --database or --collection

source/includes/options-mongorestore.yaml
```diff
@@ -239,6 +240,12 @@ description: |
   level of the dump directory for :program:`mongorestore --oplogReplay
   <mongorestore>` to replay.
 
+  :program:`mongorestore --oplogReplay <mongorestore>` must restore a
+  full dump of a :term:`replica set` member created using 
+  :program:`mongodump --oplog <mongodump>`. You cannot use 
+  :option:`--database` or :option:`--collection` with {{role}} to
```
rkumar-mongo: Are these the only two options? There are a number of include/exclude options related to namespaces as well - does --oplogReplay work with those?

If so, rather than having a full list I think you could make this generic:

"``--oplogReplay`` restores the entire dump, and is incompatible with limiter options such as --database, --collection, or --nsExclude"

SpencerBrown:
Yes, using an option that limits the restore scope with `--oplogreplay` will not work and causes an error message.

SpencerBrown:
Given that we are updating the docs for `mongorestore` to clarify that a full restore is required for `--oplogReplay`....

we also need to update the docs for `mongodump` as per DOCS-10705, to give similar caveats for `mongodump` with the `--oplog` option.

atsansone:
@SpencerBrown : I added everything for DOCS-10705. Can I get your review / LGTM?

SpencerBrown:
the --nsinclude and --nsexclude options are also mutually exclusive with --oplogReplay. example:

```            
Spencer-Brown:tls-certs spencer$ mongorestore --oplogReplay --nsInclude=foo
2017-11-17T16:50:17.398-0600    Failed: cannot use --oplogReplay with includes specified
Spencer-Brown:tls-certs spencer$ mongorestore --version
mongorestore version: r3.4.10
git version: 078f28920cb24de0dd479b5ea6c66c644f6326e9
Go version: go1.7
   os: darwin
   arch: amd64
   compiler: gc
OpenSSL version: OpenSSL 0.9.8zh 14 Jan 2016
Spencer-Brown:tls-certs spencer$
```

atsansone:
@kay-kim : RFM for `master` and `v3.4`.

source/includes/options-mongodump.yaml
```diff
@@ -252,6 +252,19 @@ description: |
   :term:`master` nodes in master/slave replication deployments.
 
   {{role}} does not dump the oplog collection.
+
+  .. note::
+
+     To use :program:`mongodump` with {{role}}, you must create a full
+     dump of a :term:`replica set` member. :program:`mongodump` with 
+     {{role}} fails if you use any of the following options along with 
+     {{role}} to limit the data to be dumped:
```
kay-kim: You already stated that mongodump with {{role}} fails, so you don't need to end the sentence with along with {{role}}

Note also, that the along with {{role}} at the end makes it sound like you are using oplogReplay to also limit the data to be dumped.

source/includes/options-mongorestore.yaml
```diff
@@ -239,8 +240,24 @@ description: |
   level of the dump directory for :program:`mongorestore --oplogReplay
   <mongorestore>` to replay.
 
+  .. note::
+
+     When using :program:`mongorestore` with {{role}}, you must 
+     restore a full dump of a :term:`replica set` member created 
+     using :program:`mongodump --oplog <mongodump>`. 
+     :program:`mongorestore` with {{role}} fails if you use any of 
+     the following options along with {{role}} to limit the data be
```
kay-kim: same

source/includes/options-mongodump.yaml
atsansone: Argh. Fixed.

source/includes/options-mongorestore.yaml
atsansone: Fixed.

PR #3063, opened 2017-10-30
DOCS-5887: mongotop --locks is only supported for 2.6 and older instances

stennie:
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/3063)


kay-kim:
merged!

PR #3072, opened 2017-11-02
Small clarification to avoid implying that the name of the primary sh…

techwriter13:
…ard is only stored when sharding it enabled.

---
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/3072)

### DIFF
--- a/source/reference/config-database.txt
+++ b/source/reference/config-database.txt
@@ -224,10 +224,10 @@ Collections
    .. include:: /includes/admonition-config-db-is-internal.rst
 
    The :data:`~config.databases` collection stores a document for each
-   database in the cluster, and tracks if the database has sharding
-   enabled. :data:`~config.databases` represents each database in a
-   distinct document. When a databases have sharding enabled, the
-   ``primary`` field holds the name of the :term:`primary shard`.
+   database in the cluster and tracks whether the database has sharding
+   enabled. The ``primary`` field holds the name of the :term:`primary
+   shard`. The ``partitioned`` field is ``true`` if the database has
+   sharding enabled.
 
    .. code-block:: javascript
 


techwriter13:
Ready for Merge

source/reference/config-database.txt
```diff
@@ -224,10 +224,10 @@ Collections
    .. include:: /includes/admonition-config-db-is-internal.rst
 
    The :data:`~config.databases` collection stores a document for each
-   database in the cluster, and tracks if the database has sharding
-   enabled. :data:`~config.databases` represents each database in a
-   distinct document. When a databases have sharding enabled, the
-   ``primary`` field holds the name of the :term:`primary shard`.
+   database in the cluster and tracks whether the database has sharding
+   enabled. The ``primary`` field holds the name of the :term:`primary
```
rkumar-mongo: cc: @jeff-allen-mongo 

While the information here is strictly correct, I think we can do a better job of making the relationship between these two fields more explicit. That would better (in my opinion) address the ticket filer's confusion regarding the role of either of these fields in relation to sharding:

This is spitballing, please clean this up :)

"The ``partitioned`` field indicates whether the database has sharding enabled. If ``partitioned`` is false, then the ``primary`` shard stores all collections for that database. If ``partitioned`` is true, the ``primary`` shard stores all unsharded collections for that database. The ``primary`` shard still behaves as a normal shard in the cluster otherwise, storing sharded collection data as needed." 

The key here is being very clear as to why the primary shard matters and its relationship to the ``partitioned`` field. The current copy defines what each of those are, but not how they relate to each other or why the ``primary`` field matters.

kay-kim:
closing as ticket has been reassigned and is in progress.

PR #3088, opened 2017-11-07
Forgot to add the link to the Journaliing article.

techwriter13:
Clarified the difference between the behavior of WiredTiger and MMAPv1 when journaling is turned on.

### DIFF
--- a/source/reference/write-concern.txt
+++ b/source/reference/write-concern.txt
@@ -158,6 +158,12 @@ the write operation has been written to the :doc:`journal
      configuration setting determines the behavior. See
      :ref:`wc-ack-behavior` for details.
 
+   - The :doc:`WiredTiger </core/wiredtiger>` and :doc:`MMAPv1 </core/mmapv1>`
+     storage engines handle setting a write concern that includes ``j: true``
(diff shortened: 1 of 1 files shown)


source/reference/write-concern.txt
```diff
@@ -158,6 +158,12 @@ the write operation has been written to the :doc:`journal
      configuration setting determines the behavior. See
      :ref:`wc-ack-behavior` for details.
 
+   - The :doc:`WiredTiger </core/wiredtiger>` and :doc:`MMAPv1 </core/mmapv1>`
+     storage engines handle setting a write concern that includes ``j: true``
```
kay-kim: This sentence makes it sound like storage engines set a write concern. 

I think we can just simplify and get to the meat:

- For a write concern that includes ``j: true``,

  - :doc:`WiredTiger </core/wiredtiger>` forces an immediate, synchronous journal flush.
  - :doc:`MMAPv1 </core/mmapv1>`  waits for the next scheduled journal flush.

kay-kim: If you could make the aforementioned changes

PR #3099, opened 2017-11-20
DOCS-11037 Describe $out stage process

kallimachos:
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/3099)

### DIFF
--- a/source/reference/operator/aggregation/out.txt
+++ b/source/reference/operator/aggregation/out.txt
@@ -25,7 +25,7 @@ Definition
    lets the aggregation framework return result sets of any size.
 
    .. versionchanged:: 3.2.0
-   
+
       MongoDB 3.2 added support for :ref:`document validation
       <3.2-rel-notes-document-validation>`. The ``bypassDocumentValidation``
       field enables you to bypass document validation during the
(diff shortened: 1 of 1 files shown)


source/reference/operator/aggregation/out.txt
```diff
@@ -68,10 +68,18 @@ Replace Existing Collection
...
-results collection. The :pipeline:`$out` operation does not change any
-indexes that existed on the previous collection. If the aggregation
-fails, the :pipeline:`$out` operation makes no changes to the pre-existing
-collection.
+results collection using the following process:
+
+#. A temp collection is created.
+#. Indexes are copied from the destination collection to the temp collection.
+#. Documents are inserted into the temp collection.
+#. The temp collection calls `renameCollection
```
kay-kim: ? the colleciton calls the renameCollection method?

Let's make it more `$out` centric:
... replaces the existing collection with the new results collection.  Specifically, the :pipeline:`$out` operation
#. Creates a temp collection.
#. Copies the indexes from the existing collection to the temp collection.
#. Inserts the documents into the temp collection.
#. Calls :method:`db.collection.renameCollection` with ``dropTarget: true`` to rename the temp collection to the destination collection.

kallimachos: That's definitely better, thanks @kay-kim.

PR #3110, opened 2017-11-22
Update configure-a-hidden-replica-set-member.txt

user501254:
The use of letter 'a' is not grammatically correct here.

---
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/3110)


jdestefano-mongo:
Thanks!

PR #3111, opened 2017-11-22
Update configure-a-non-voting-replica-set-member.txt

user501254:
The use of letter 'an' is not grammatically correct here.

---
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/3111)


jdestefano-mongo:
Thanks!

PR #3117, opened 2017-11-24
DOCS-11052: Improve Monitor MongoDB with SNMP on Linux page

kallimachos:


emisca:
please check my comment on the DOCS ticket before going ahead. This needs to be reworked.

kallimachos:
Ack. Thanks @emisca.

emisca:
Hi @kallimachos ,

I see that you are now providing instructions on how to configure snmpd as a master on /etc/snmp/snmpd.conf. 

Make sure to clarify that this is only an example and will work only with the Net-SNMP snmpd daemon. Customers should refer to their Linux distribution documentation and eventually contact their support for further guidance on how to configure it. Make clear that this is out of scope for MongoDB support and we cover only the configuration on the Mongodb side.

PR #3154, opened 2017-12-07
fix typo: slowly should be quickly

lutaoact:
so I need to balance

---
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/3154)

### DIFF
--- a/source/tutorial/manage-sharded-cluster-balancer.txt
+++ b/source/tutorial/manage-sharded-cluster-balancer.txt
@@ -99,7 +99,7 @@ To configure default chunk size, see
 Schedule the Balancing Window
 -----------------------------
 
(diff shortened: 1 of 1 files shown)


p-mongo:
I think the idea with the existing language is that balancing can be postponed (to the next window) if the write activity is low. If the write activity is high balancing may need to be run more frequently.

kay-kim:
Hi. Thanks for this pull request and apologies for the long delay.

As noted above, the intention was to convey a low rate of write activity - and as such, I believe that slowly is correct (although, we could definitely have written it better).  

As such, will close this pull request without merge. However, we very much appreciate your taking the time to help us make the documentation better and once again, apologies for the long delay.

Regards,

Kay Kim

PR #3157, opened 2017-12-08
DOCS-8981: Document maxLogSizeKB parameter

kallimachos:
Also fix two minor typos.

---
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/3157)

### DIFF
--- a/source/reference/command/setParameter.txt
+++ b/source/reference/command/setParameter.txt
@@ -34,6 +34,7 @@ setParameter
    - :parameter:`logComponentVerbosity`
    - :parameter:`logLevel`
    - :parameter:`logUserIds`
+   - :parameter:`maxLogSizeKB`
    - :parameter:`quiet`
    - :parameter:`traceExceptions`
 
--- a/source/reference/parameters.txt
+++ b/source/reference/parameters.txt
@@ -756,6 +756,11 @@ Logging Parameters
 
       mongod --setParameter maxLogSizeKB=20
 
+   .. warning::
+
+      Using a large value for :parameter:`maxLogSizeKB` may adversely affect
+      system performance and negatively impact database operations.
+
 .. parameter:: quiet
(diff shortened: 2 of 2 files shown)


devkev:
This is good, thanks for that.  My only suggestion is to add a warning that large values of this parameter can adversely affect system performance.

As an example of the potential impact (this detail doesn't necessarily have to end up in the docs), suppose a mongod/mongos is attempting to log 100 "long" (over 10MB) lines per second (massive queries, say).  With the default maxLogSizeKB of 10, this is a logging rate of about 1MB/s.  However, with a value of 10000, it's a logging rate of about 1GB/s.  This is extra bad because log lines are written synchronously - which is why I'm concerned about it affecting the perf of actual db ops.

---

Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/mongodb/docs/3157#-:-L02-CG1IgOZ0e2q6G5H:blr6hjk)*

kallimachos:
@devkev If you would be so kind as to ignore the extraneous commits from rebasing and just have a look at `source/reference/parameters.txt` for the added warning, I'd be much obliged.

Thanks!

devkev:
:lgtm:

---

Reviewed 1 of 1 files at r3.
Review status: 2 of 5 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/mongodb/docs/3157#-:-L03TMI5w_2MMk-LsNFH:bnfp4nl)*

kay-kim:
@kallimachos -- could you do a rebase so that we don't have the other commits in this pull request?

kallimachos:
@kay-kim Should be better now. Thanks!

PR #3158, opened 2017-12-10
Took out --with-openssl option b/c doesn't exist

kunal732:
There is no available option `--with-openssl` . This is installed by default. Adding this option gives the user a warning during install.

---
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/3158)


kay-kim:
Thank you!

PR #3160, opened 2017-12-12
DOCS-8794: Document sort order for Object

sindbach:
Based on 
https://github.com/mongodb/mongo/blob/r3.4.10/src/mongo/bson/bsonobj.cpp#L105 

applies to MongoDB v3.4 and v3.6

---
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/3160)

### DIFF
--- a/source/reference/bson-type-comparison-order.txt
+++ b/source/reference/bson-type-comparison-order.txt
@@ -55,6 +55,17 @@ when comparing a field whose value is a single-element array (e.g. ``[
 ``1`` and ``2``. A comparison of an empty array (e.g. ``[ ]``) treats
 the empty array as less than ``null`` or a missing field.
 
+Objects
+-------
+
+A comparison of two BSON objects using operators is performed using the following order : 
+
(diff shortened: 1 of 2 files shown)


source/reference/bson-type-comparison-order.txt
```diff
@@ -55,6 +55,17 @@ when comparing a field whose value is a single-element array (e.g. ``[
 ``1`` and ``2``. A comparison of an empty array (e.g. ``[ ]``) treats
 the empty array as less than ``null`` or a missing field.
 
+Objects
+-------
+
```
kallimachos: .. versionadded:: 3.4

sindbach: This behaviour wasn't changed at 3.4. Not sure what was the first version the behaviour introduced, but just never documented.

kallimachos: Ah, that makes sense.

schmalliso:
Merged to master - thanks!

schmalliso:
Oh, and back-ported to 3.4 as well.

PR #3214, opened 2018-01-31
DOCS-8977: include filter param for db.getCollectionInfos()

schmalliso:
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/3214)


schmalliso:
We address the filter in the example section, but weirdly didn't reference it in the top-of-page method reference.

PR #3262, opened 2018-03-14
(DOCS-9990, -10826): Updated flushRouterConfig.

atsansone:
@steveren , @kaloianm : This PR adds the invocation example and changes the explanation as to how the cache is cleared and rebuilt. [Staged](https://docs-mongodbcom-staging.corp.mongodb.com/anthonysansone/DOCS-10826/reference/command/flushRouterConfig.html)

---
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/3262)

### DIFF
--- a/source/reference/command/flushRouterConfig.txt
+++ b/source/reference/command/flushRouterConfig.txt
@@ -13,18 +13,25 @@ flushRouterConfig
 .. dbcommand:: flushRouterConfig
 
    :dbcommand:`flushRouterConfig` clears the cached routing table in a
-   :binary:`~bin.mongos` instance. Subsequent :binary:`~bin.mongos`
-   commands update the routing table as routes are needed.
+   :binary:`~bin.mongos` instance. Subsequent commands against that 
+   :binary:`~bin.mongos` will repopulate the routing table cache from 
+   the :term:`config server` as they execute.
 
-   When this command is called, it forces an update when the
-   configuration database holds data that is more recent than the data
-   cached in the :binary:`~bin.mongos` process.
+   Use this command to force the routing table cache to be refreshed 
+   from the :term:`config server` if it contains more recent data. In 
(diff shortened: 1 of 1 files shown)


source/reference/command/flushRouterConfig.txt
```diff
@@ -12,13 +12,13 @@ flushRouterConfig
 
 .. dbcommand:: flushRouterConfig
 
-   :dbcommand:`flushRouterConfig` clears the current cluster
-   information cached by a :binary:`~bin.mongos` instance and reloads all
-   :term:`sharded cluster` metadata from the :term:`config database`.
+   :dbcommand:`flushRouterConfig` clears the cached routing table in a
+   :binary:`~bin.mongos` instance. Subsequent :binary:`~bin.mongos`
+   commands update the routing table as routes are needed.
```
kaloianm: Subsequent commands against that :binary:`~bin.mongos` will repopulate the routing table cache from the config server as they execute.

source/reference/command/flushRouterConfig.txt
```diff
@@ -12,13 +12,13 @@ flushRouterConfig
...
 
-   This forces an update when the configuration database holds data
-   that is newer than the data cached in the :binary:`~bin.mongos`
-   process.
+   When this command is called, it forces an update when the
+   configuration database holds data that is more recent than the data
+   cached in the :binary:`~bin.mongos` process.
```
kaloianm: This sounds a little awkward. How about:

Use this command to force the routing table cache to be refreshed from the config server when there is possibility that it contains a more recent data. In almost all cases, this happens automatically, so it typically only needs to be done after a movePrimary as per instructions.

atsansone:
@kaloianm : Ready for another pass. Thanks!

atsansone:
@kay-kim : RFM for `v3.6` and `master`. Thanks!

PR #3265, opened 2018-03-16
Rename authorization to authentication for --auth command line option to match --noauth and elsewhere

reddog:
Compare https://docs.mongodb.com/manual/reference/program/mongod/#cmdoption-mongod-auth

> --auth
> Enables **authorization** to control user’s access to database resources and operations. When **authorization** is enabled, MongoDB requires all clients to authenticate themselves first in order to determine the access for the client.

with https://docs.mongodb.com/manual/reference/program/mongod/#cmdoption-mongod-noauth

> --noauth
> Disables **authentication**. Currently the default. Exists for future compatibility and clarity.

---
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/3265)

### DIFF
--- a/locale/es/LC_MESSAGES/reference/program/mongod.po
+++ b/locale/es/LC_MESSAGES/reference/program/mongod.po
@@ -1471,7 +1471,7 @@ msgstr ""
 
 #: ../source/includes/option/option-mongod-auth.rst:3
 msgid ""
-"Enables authorization to control user's access to database resources and "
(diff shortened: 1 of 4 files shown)


reddog:
I got the terminology change the wrong way around!  It *should* be `authorization`, see https://docs.mongodb.com/manual/reference/configuration-options/#security.authorization  This means that the description for `--noauth` is the one that needs changing

PR #3266, opened 2018-03-17
Update read-concern.txt

shannonbradshaw:
Our discussion of afterClusterTime here:
https://docs.mongodb.com/manual/reference/read-concern/#afterclustertime
does not currently explain exactly what effect afterClusterTime has on the way an instance handles operations. 

In addition, we don't explain readConcern: available which leaves the reader at a bit of a cliff hanger.

Finally, there are several grammatical mistakes.

---
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/3266)

### DIFF
--- a/source/reference/read-concern.txt
+++ b/source/reference/read-concern.txt
@@ -136,16 +136,16 @@ The following read concern levels are available:
 
 .. versionadded:: 3.6
 
-MongoDB introduces the ``afterClusterTime`` option that can be set by
-the drivers.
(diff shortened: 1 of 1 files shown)


kay-kim:
@shannonbradshaw  --- why do you not watch bad movies on your flights?  Will merge and tweak some rst things so that we link properly.

shannonbradshaw:
:-) Thanks!

PR #3276, opened 2018-03-30
DOCSP-2133: Clarify time required for failover/election

rkumar-mongo:
@kay-kim ready for merge into 3.6 branch. Let me know if this looks OK. There's a lot of green but the majority is merged up from the elections page. Goal was to make the Automatic Failover section + High Availability section very easy to follow and set expectations for the typical period of time where a cluster may not have a primary.


source/core/replica-set-high-availability.txt
```diff
@@ -15,29 +19,242 @@ Replica Set High Availability
...
+
+If a majority of replica set members are inaccessible or unavailable, 
+the remaining members of the replica set cannot hold elections.
+The primary node steps down until a majority of members are 
+online and the cluster becomes read-only. The cluster can continue to 
+serve read queries if such queries are configured to run on secondaries. 
+See :ref:`replica-set-read-preference` for complete documentation on 
+secondary read operations.
+
+Factors and Conditions that Affect Elections
```
rkumar-mongo: All content below this section is pulled in from the Elections page and is out of scope for this review.

source/core/replica-set-high-availability.txt
```diff
@@ -15,29 +19,242 @@ Replica Set High Availability
...
+continue to interact with the cluster during elections, mitigating
+the risk of downtime due to critical errors such as network 
+disruptions and hardware failure.
+
+Replica sets can trigger an election in response to a variety of events,
+such as:
+
+- :method:`initiating a replica set <rs.initiate()>`,
+- performing replica set maintenance using methods such as :method:`rs.stepDown()` or :method:`rs.reconfig()`, and
+- losing connectivity to the primary for more than the configured :rsconf:`timeout <settings.electionTimeoutMillis>` (10 seconds by default).
```
stbrody: who's losing connectivity?
Probably makes sense to add "A node in the cluster" to the beginning of this bullet

source/core/replica-set-high-availability.txt
```diff
@@ -15,29 +19,242 @@ Replica Set High Availability
...
+
+In the following diagram, the primary node unavailability
+exceeded the 
+:rsconf:`configured timeout <settings.electionTimeoutMillis>`.
+One of the remaining secondaries holds an election in response to
```
stbrody: s/holds/calls for.  everyone's involved in the election, just one node initiated it

source/core/replica-set-high-availability.txt
```diff
@@ -15,29 +19,242 @@ Replica Set High Availability
...
+:rsconf:`protocolVersion: 1 <protocolVersion>`. For elections
+triggered by primary failover, the total time in which
+a cluster may operate without a primary should not typically exceed
+12 seconds assuming the default settings for automatic failover.
+See :ref:`replication-auto-failover` for more information.
+
+.. include:: /includes/fact-election-latency.rst
+
+If a majority of replica set members are inaccessible or unavailable, 
+the remaining members of the replica set cannot hold elections.
```
stbrody: They can hold elections, they just can't win one.

source/core/replica-set-high-availability.txt
```diff
@@ -15,29 +19,242 @@ Replica Set High Availability
...
+successfully. 
+
+.. include:: /includes/fact-retryable-writes-election.rst
+
+Since MongoDB 3.2, elections typically complete within two seconds
+in clusters running with the default 
+:rsconf:`protocolVersion: 1 <protocolVersion>`. For elections
+triggered by primary failover, the total time in which
+a cluster may operate without a primary should not typically exceed
+12 seconds assuming the default settings for automatic failover.
```
stbrody: I'd be more comfortable if we said 15 seconds here, to cover us in the event that the first election results in a split vote and a second election needs to be run.

source/reference/command/replSetReconfig.txt
```diff
@@ -53,15 +53,25 @@ considerations:
...
-  members of the replica set. When this happens, the primary node will drop
-  all current connections.
+- The :method:`rs.reconfig()` shell method can trigger the current 
+  primary to step down in some situations. When the primary steps down, 
+  it forcibly closes all client connections.
+  
+  If this command steps down the current primary, it also triggers an 
+  :ref:`election <replica-set-elections>` to select a new 
+  :term:`primary`. The election process typically completes 
+  within 2 seconds. See :ref:`replica-set-elections` for more
```
stbrody: it should complete within 2 seconds once the election is called for, but it can take up to electionTimeoutMillis for the election to start.  Reconfig doesn't actually force an election to happen proactively, it just causes the primary to step down, which will result in an election once the secondaries notice the lack of a primary

source/replication.txt
```diff
@@ -104,30 +104,43 @@ Automatic Failover
 ------------------
 
 When a primary does not communicate with the other members of the set
-for more than 10 seconds, an eligible secondary will hold an election
+for more than 10 seconds, an eligible secondary holds an election
```
stbrody: electionTimeoutMillis seconds, which happens to default to 10

source/replication.txt
```diff
@@ -104,30 +104,43 @@ Automatic Failover
...
+
+.. include:: /includes/fact-election-latency.rst
+
+See :ref:`replica-set-elections` for complete documentation on
+replica set elections.  
+
+When the former primary recovers from a failover event and reconnects to 
+the cluster, MongoDB may rollback any write operations that were *not*
+replicated to secondary members at the time of failure. MongoDB attempts
+to avoid rollbacks, which should be rare.
```
stbrody: I think this sends the wrong message.  We shouldn't be promising that rollbacks are rare, we should be emphasizing that any data written and confirmed with w:majority won't roll back

source/core/replica-set-high-availability.txt
stbrody: I assume you mean only in this one file?  Other files below this seem to have substantive changes, and so I've gone ahead and reviewed them.

source/core/replica-set-high-availability.txt
rkumar-mongo: Done and Done.

source/core/replica-set-high-availability.txt
rkumar-mongo: Fixed. also made a tweak, as the sentence as is implies the secondary that calls for an election is the one that would win. Unless its technically accurate to say that.

source/core/replica-set-high-availability.txt
rkumar-mongo: Done.

source/core/replica-set-high-availability.txt
rkumar-mongo: Done. I tweaked this paragraph a bit in general as well.

source/core/replica-set-high-availability.txt
rkumar-mongo: That's correct, thank you.

source/reference/command/replSetReconfig.txt
rkumar-mongo: Made some tweaks to emphasize this.

source/replication.txt
rkumar-mongo: Done, with some tweaks

source/replication.txt
rkumar-mongo: Done

source/replication.txt
```diff
@@ -104,30 +104,43 @@ Automatic Failover
...
-   enhancements <3.2-rel-notes-rs-enhancements>` for details.
+.. include:: /includes/fact-retryable-writes-failover-election.rst 
 
-See :ref:`replica-set-elections` and
-:ref:`replica-set-rollbacks` for more information.
+Since MongoDB 3.2, elections typically complete within two seconds
+in clusters running with the default 
+:rsconf:`protocolVersion: 1 <protocolVersion>` . The total 
+time in which a cluster may operate without a primary should not 
+typically exceed 12 seconds.
```
shannonbradshaw: "The total time in which a cluster may operate without a primary should not typically exceed 12 seconds."

This sentence nullifies the sentence before it and leaves me confused as to whether 2 or 12 seconds is typical. The paragraph that follows this in the rendered view, which comes from the include fact-election-latency.rst provides sufficient hinting that elections might take longer in some circumstances.

source/replication.txt
```diff
@@ -138,8 +139,9 @@ replica set elections.
 
 When the former primary recovers from a failover event and reconnects to 
 the cluster, MongoDB may rollback any write operations that were *not*
-replicated to secondary members at the time of failure. MongoDB attempts
-to avoid rollbacks, which should be rare. 
+replicated to secondary members at the time of failure. Write operations
+issued with a :ref:`write concern <write-concern>` of 
+:writeconcern:`majority` that return successfully do not roll back.
 See :ref:`replica-set-rollbacks` for more information.
```
shannonbradshaw: I suggest striking this paragraph entirely. It's difficult to say anything meaningful here without either sending the message that rollbacks are common or that all writes should use w: majority. Addressing either issue is too in the weeds for this overview page. Instead, I suggest a simple statement pointing readers to further information about writes in replica sets. Something like, "To learn more about how writes are handled in replica sets during failover scenarios see:" and then point the reader to:
- reference/write-concern.txt
- core/replica-set-rollbacks.txt

source/replication.txt
rkumar-mongo: I'm wondering if we should frame this in reverse.

e.g., "The total time in which a cluster may operate without a primary should not typically exceed 15 seconds, and may resolve as quickly as within two seconds."

In a sense that has the same issue - I do think there is some value in setting a 'range' that users can generally expect. Alternatively, we can just stick to the 15-second upper-limit and let users be pleasantly surprised whenever its faster. @stbrody thoughts on the latter?

shannonbradshaw: One primary motivation for this ticket is that customers are raising concerns about the amount of time failover takes as reported by our docs and that this can cause friction in the sales process. I think we should report a middle number. The median is reasonable. In the ticket description Cailin suggests this is 10.5 - 11.5 seconds (with default settings) and only 5.5-6.5 seconds in Atlas.

source/replication.txt
rkumar-mongo: Done

source/core/replica-set-high-availability.txt
```diff
@@ -15,29 +19,244 @@ Replica Set High Availability
    :depth: 1
    :class: singlecol
 
-:term:`Replica sets <replica set>` provide high availability using
-automatic :term:`failover`. Failover allows a :term:`secondary` member
-to become :term:`primary` if the current primary becomes unavailable.
+:term:`Replica sets <replica set>` use elections to select the 
+:term:`primary` for the cluster. Replica set elections support
```
ntang: I would use "primary node" or "primary member", it seems more clear to me, unless that violates other examples.

source/core/replica-set-high-availability.txt
```diff
@@ -15,29 +19,244 @@ Replica Set High Availability
...
+high availability by allowing clusters to automatically recover
```
ntang: I'd use "resume" and not "recover".  The app can recover from the failover, the cluster can resume operations.

source/core/replica-set-rollbacks.txt
```diff
@@ -100,4 +100,4 @@ the set by deleting the content of the :setting:`~storage.dbPath` directory for
 the member that requires a larger rollback.
 
 .. seealso:: :doc:`/core/replica-set-high-availability` and
-   :doc:`/core/replica-set-elections`.
+   :doc:`/core/replica-set-high-availability`.
```
ntang: If I'm reading this correctly, it seems to say "see also replica-set-high-availability and replica-set-high-availability" which seems redundant.

source/core/replica-set-high-availability.txt
ntang: Also, it feels like "Replica sets support high availability by holding elections..." would make more sense.  I guess technically the election is supporting it too, but it feels more like the election and subsequent resumption of normal operations is _how_ replica sets support high availability.

source/core/replica-set-high-availability.txt
ntang: Or you could say the cluster recovers from a fault in a node, the network, etc. and then resumes normal operations.  But recovering normal operations implies that it will do something to make sure that any normal operations that were underway will still succeed; it won't.  That's up to the app and its retry logic and/ or write queuing etc. to handle.

source/core/replica-set-architecture-three-members.txt
```diff
@@ -27,7 +27,7 @@ A replica set with three members that store data has:
 
 - Two :doc:`secondary </core/replica-set-secondary>` members. Both
   secondaries can become the primary in an :doc:`election
-  </core/replica-set-elections>`.
+  </core/replica-set-high-availability>`.
```
ntang: I know I'm pedantic but I really don't like renaming this url to high availability.  I feel like replica-set-availability would be a much better name; high availability is what you get from using the availability features correctly.

source/core/replica-set-high-availability.txt
```diff
@@ -15,29 +19,244 @@ Replica Set High Availability
...
    :class: singlecol
 
-:term:`Replica sets <replica set>` provide high availability using
-automatic :term:`failover`. Failover allows a :term:`secondary` member
-to become :term:`primary` if the current primary becomes unavailable.
+:term:`Replica sets <replica set>` use elections to select the 
+:term:`primary` for the cluster. Replica set elections support
+high availability by allowing clusters to automatically recover 
+normal operations after a primary failover event. Applications can
+continue to interact with the cluster during elections, mitigating
```
ntang: AFAIK the only interactions allowed are read operations.  (I may also not be thinking of something.)  If this is the case, I'd change "interact with" to "read from".  I don't think we should hide the fact that writes don't get processed during the election; that will set incorrect expectations and those will then lead to customers and users getting pissed off and saying our docs are wrong when they have an election and writes stop.  Being able to read during an election still does mitigate the risk of downtime.  Maybe it's also worth adding a note that writes can be queued and/ or retried from the application side to minimize the impact of the election.

source/core/replica-set-high-availability.txt
```diff
@@ -15,29 +19,244 @@ Replica Set High Availability
...
+such as:
+
+- Adding a new node to the replica set,
+- :method:`initiating a replica set <rs.initiate()>`,
+- performing replica set maintenance using methods such as :method:`rs.stepDown()` or :method:`rs.reconfig()`, and
+- the :term:`secondary` members losing connectivity to the primary for more than the configured :rsconf:`timeout <settings.electionTimeoutMillis>` (10 seconds by default).
+
+In the following diagram, the primary node unavailability
+exceeded the 
+:rsconf:`configured timeout <settings.electionTimeoutMillis>`.
```
kay-kim: Could we tie it a bit closer to the last bullet point on line 36?

source/core/replica-set-high-availability.txt
```diff
@@ -15,29 +19,244 @@ Replica Set High Availability
...
+
+.. include:: /images/replica-set-trigger-election.rst
+
+During an election the replica set can continue to serve read queries 
+if such queries are configured to 
+:ref:`run on secondaries <replica-set-read-preference>`. The
+replica set cannot process write operations until the election completes
+successfully. 
+
+.. include:: /includes/fact-retryable-writes-election.rst
```
kay-kim: Starting with this include, they just seem just a collection of random facts.  We can organize a bit better.

source/core/replica-set-high-availability.txt
```diff
@@ -15,29 +19,244 @@ Replica Set High Availability
...
+The median time in which a cluster may operate without a primary
+should not typically exceed 11.5 seconds, assuming default
+:rsconf:`replica configuration settings <settings>`. 
+This includes time required to mark the primary as
+:ref:`unavailable <replication-auto-failover>`.
+
+.. include:: /includes/fact-election-latency.rst
+
+If a majority of replica set members are inaccessible or unavailable,
+the remaining minority of replica set membrers cannot meet quorum for
```
kay-kim: typo -- membrers

source/core/replica-set-high-availability.txt
```diff
@@ -15,29 +19,244 @@ Replica Set High Availability
...
+electing a new primary. If the primary node is among the remaining
+minority of replica set, it steps down and the cluster becomes 
+read-only. When a majority of replica set members are available, the 
+members can meet quurom for electing a new primary. Until that time, 
+the cluster can continue to serve read queries if such queries are 
+configured to run on secondaries. See :ref:`replica-set-read-preference`
+for complete documentation on secondary read operations.
```
kay-kim: I think we can tighten this paragraph a bit, but may be taken care of if we tighten up the presentation of information a bit more on this page.

Also, isn't this kind of similar to the information regarding network partition?

source/includes/fact-retryable-writes-election.rst
```diff
@@ -0,0 +1,10 @@
+Your application connection logic should include tolerance for 
+elections, such as queueing write operations until it detects a healthy 
+primary.
```
kay-kim: Can you explain what you mean by queuing write operations until it detects a healthy primary? and how that interacts with the retryable writes?

source/includes/fact-retryable-writes-failover-election.rst
```diff
@@ -0,0 +1,10 @@
+Your application connection logic should include tolerance for
```
kay-kim: Why do we have two includes that say the same thing?  Doesn't it defeat the purpose of single sourcing?

source/reference/command/replSetReconfig.txt
```diff
@@ -53,15 +53,28 @@ considerations:
...
-- This command can cause downtime as the set renegotiates
-  primary-status. Typically this is 10-20 seconds, but could
-  be as long as a minute or more. Therefore, you should attempt
-  to reconfigure only during scheduled maintenance periods.
-
-- In some cases, :dbcommand:`replSetReconfig` forces the current
-  primary to step down, initiating an election for primary among the
-  members of the replica set. When this happens, the primary node will drop
-  all current connections.
+- The :method:`rs.reconfig()` shell method can trigger the current
```
kay-kim: This bullet is in the replSetReconfig page. This makes it sound like if you run the command replSetReconfig, no primary stepdown, whereas if you use the shell method, well, hello stepdown.

source/reference/command/replSetReconfig.txt
```diff
@@ -53,15 +53,28 @@ considerations:
...
-  primary to step down, initiating an election for primary among the
-  members of the replica set. When this happens, the primary node will drop
-  all current connections.
+- The :method:`rs.reconfig()` shell method can trigger the current 
+  primary to step down in some situations. When the primary steps down, 
+  it forcibly closes all client connections.
+  
+  Primary step-down triggers an :ref:`election <replica-set-elections>` 
+  to select a new :term:`primary`. The median time in which a cluster 
+  may operate without a primary should not typically exceed 11.5
```
kay-kim: "in which a cluster may operate without a primary"  makes it sound like the replica set  can still work without a primary for about 11.5 seconds but after 11.5 seconds, well...

source/reference/method/rs.reconfig.txt
```diff
@@ -46,10 +46,27 @@ Availability
...
-to step down in some situations. When the
-primary steps down, it forcibly closes all client connections. This is
-by design. Since it may take a period of time to elect a new primary, 
-schedule reconfiguration changes during maintenance periods to minimize loss of write availability.
+to step down in some situations. When the primary steps down, it 
+forcibly closes all client connections.
+
+Primary step-down triggers an :ref:`election <replica-set-elections>` 
+to select a new :term:`primary`. The median time in which a cluster 
+may operate without a primary should not typically exceed 11.5
```
kay-kim: same comment as above.

source/replication.txt
```diff
@@ -104,30 +104,45 @@ Automatic Failover
...
-Although the timing varies, the failover process generally completes
-within a minute. For instance, it may take 10-30 seconds for the
-members of a :term:`replica set` to declare a :term:`primary`
-inaccessible (see :rsconf:`~settings.electionTimeoutMillis`). One of
-the remaining secondaries holds an :term:`election` to elect itself as
-a new primary. The election itself may take another 10-30 seconds.
+Lowering the :rsconf:`~settings.electionTimeoutMillis` 
+replication configuration option from the default ``10000`` (10 seconds)
+can result in faster automated failover at the cost of increased
+frequency of elections.
```
kay-kim: When I read that, I think oh, more elections but faster failover (which includes the election process) so, possibly my outage could be equivalent. 

we should mention the impact on w:1 writes.

source/replication.txt
```diff
@@ -104,30 +104,45 @@ Automatic Failover
...
 
-   Starting in MongoDB 3.2, with the :ref:`replication election
-   enhancements <3.2-rel-notes-rs-enhancements>`, MongoDB reduces
-   replica set failover time. See :ref:`replication election
-   enhancements <3.2-rel-notes-rs-enhancements>` for details.
+.. include:: /includes/fact-retryable-writes-failover-election.rst 
 
-See :ref:`replica-set-elections` and
-:ref:`replica-set-rollbacks` for more information.
+The median time in which a cluster may operate without a primary
```
kay-kim: Same comment as in other file. But also, should we mention median time and time before we talk about doing things to reduce the timing?

source/core/replica-set-high-availability.txt
stbrody: When a node calls for an election, it is fundamentally nominating itself.  But if multiple nodes call for an election at the same time, there's no guarantee which (if any) of them wins.  I think the way you reworded it is fine

source/core/replica-set-high-availability.txt
```diff
@@ -15,29 +19,226 @@ Replica Set High Availability
...
+
+The replica set cannot process write operations until the
+election completes successfully. The replica set can continue to serve 
+read queries if such queries are configured to 
+:ref:`run on secondaries <replica-set-read-preference>`.
+
+.. include:: /includes/fact-retryable-writes-failover-election.rst
+
+The median time before a cluster elects a new primary should not
+typically exceed 11.5 seconds, assuming default :rsconf:`replica
```
stbrody: saying 11.5 seems super precise given this number is actually not at all precise.  I'd prefer we say 15, or even 12.  Saying anything to a 10th of a second precision suggests this is a specific number we enforce or test.

source/core/replica-set-high-availability.txt
```diff
@@ -15,29 +19,226 @@ Replica Set High Availability
...
+configuration settings <settings>`. This includes time required to
+mark the primary as :ref:`unavailable <replication-auto-failover>` and
```
stbrody: Maybe add a sentence along the lines of "the former can be affected by changing the electionTimeoutMillis replica set config option"?

source/core/replica-set-high-availability.txt
rkumar-mongo: Fixed, thanks.

source/core/replica-set-high-availability.txt
rkumar-mongo: Done

source/core/replica-set-high-availability.txt
rkumar-mongo: Done

source/replication.txt
```diff
@@ -104,30 +104,53 @@ Automatic Failover
...
-See :ref:`replica-set-elections` and
-:ref:`replica-set-rollbacks` for more information.
+.. include:: /includes/fact-election-latency.rst
+
+Lowering the :rsconf:`~settings.electionTimeoutMillis` 
+replication configuration option from the default ``10000`` (10 seconds)
+can result in faster detection of primary failure. However,
+the cluster may call elections more frequently due to factors such as
+temporary network latency even if the primary is otherwise healthy.
+This can also result in increased
```
kay-kim: I'd remove the also.

rkumar-mongo: Done

source/replication.txt
```diff
@@ -104,30 +104,53 @@ Automatic Failover
...
+configuration settings <settings>`. This includes time required to
+mark the primary as :ref:`unavailable <replication-auto-failover>` and
+call and complete an :ref:`election <replica-set-elections>`.
+You can tune this time period by modifying the 
+:rsconf:`settings.electionTimeoutMillis` replication configuration 
+option.
 
-See :ref:`replica-set-elections` and
-:ref:`replica-set-rollbacks` for more information.
+.. include:: /includes/fact-election-latency.rst
```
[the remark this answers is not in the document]
rkumar-mongo: Added the previous paragraph to the include and merged. The two are basically linked anyways.

source/replication.txt
```diff
@@ -104,30 +104,53 @@ Automatic Failover
...
 
 When a primary does not communicate with the other members of the set
-for more than 10 seconds, an eligible secondary will hold an election
-to elect itself the new primary. The first secondary to hold an
-election and receive a majority of the members' votes becomes primary.
+for more than the configured :rsconf:`electionTimeoutMillis` period 
+(10 seconds by default), an eligible secondary calls for an election
+to nominate itself as the new primary. The cluster attempts to 
+complete the election of a new primary and ensure continued write 
+availability.
```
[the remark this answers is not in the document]
rkumar-mongo: Correct. revised.

source/replication.txt
```diff
@@ -104,30 +104,53 @@ Automatic Failover
...
 
 .. include:: /includes/fact-replica-set-protocolVersion1.rst
```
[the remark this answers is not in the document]
rkumar-mongo: Good point. removed

source/replication.txt
```diff
@@ -104,30 +104,53 @@ Automatic Failover
...
 
 .. include:: /images/replica-set-trigger-election.rst
 
-Although the timing varies, the failover process generally completes
-within a minute. For instance, it may take 10-30 seconds for the
-members of a :term:`replica set` to declare a :term:`primary`
-inaccessible (see :rsconf:`~settings.electionTimeoutMillis`). One of
-the remaining secondaries holds an :term:`election` to elect itself as
-a new primary. The election itself may take another 10-30 seconds.
+The replica set can continue to serve read queries
```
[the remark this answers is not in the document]
rkumar-mongo: Going with 1, 2, 3, 6, 7, 4, 5.

rkumar-mongo:
Closing this PR, will re-open with a fresh branch.

PR #3302, opened 2018-04-24
DOCS-11651: change cm.max_ap_count for linux

skerschb:


p-mongo:
@skerschb Could you please elaborate on why 128000 was insufficient and 135000 is?

https://jira.mongodb.org/browse/DOCS-11161 does not appear to provide any help with this.

PR #3304, opened 2018-04-30
Update and rename findOneAndDelete to findOneAndRemove

yuriy-vasilyev:
Replace "findOneAndDelete" method name with "findOneAndRemove" which is how it really goes in the code

---
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/3304)

### DIFF
--- a/source/reference/method/db.collection.findOneAndRemove.txt
+++ b/source/reference/method/db.collection.findOneAndRemove.txt
@@ -1,5 +1,5 @@
 =================================
-db.collection.findOneAndDelete()
+db.collection.findOneAndRemove()
 =================================
 
 .. default-domain:: mongodb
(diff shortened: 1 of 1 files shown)


kay-kim:
Hi -- 
thanks for your update.  The documentation is for the mongo shell helper method -- findOneAndDelete:

MongoDB Enterprise mongos> db.c.findOneAndDelete
function (filter, options) {
    var opts = Object.extend({}, options || {});
    // Set up the command
    var cmd = {query: filter, remove: true};

    if (opts.sort) {
        cmd.sort = opts.sort;

...
As such, I will go ahead and close this pull without merging.  But thanks again for taking the time.

Regards,

Kay Kim

PR #3306, opened 2018-04-30
DOCSP-2214: Clarify SASL mechanism support, highlight GSSAPI

rkumar-mongo:
@kay-kim this is RFM for 3.6, and I believe we can safely backport to 3.4 as well.


rkumar-mongo:
@kay-kim ping!

PR #3328, opened 2018-06-05
(DOCS-11026): Fixed version numbers for directory paths.

atsansone:
@kay-kim : RFM for `master` and `v3.6`. Also fixed an error on the time zone page that broke a build.

---
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs/3328)

### DIFF
--- a/source/includes/steps-configure-windows-service-for-mongodb.yaml
+++ b/source/includes/steps-configure-windows-service-for-mongodb.yaml
@@ -29,7 +29,7 @@ action:
     language: powershell
     copyable: true
     code: |
-      "C:\Program Files\MongoDB\Server\3.6\bin\mongod.exe" --config "C:\Program Files\MongoDB\Server\3.6\mongod.cfg" --install
+      "C:\Program Files\MongoDB\Server\{+version+}\bin\mongod.exe" --config "C:\Program Files\MongoDB\Server\{+version+}\mongod.cfg" --install
 post: |
   To use an alternate :setting:`dbpath`, specify the path in the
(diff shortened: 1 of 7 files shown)


source/includes/steps-configure-windows-service-for-mongodb.yaml
```diff
@@ -67,7 +67,7 @@ title: Connect to MongoDB.
 stepnum: 7
 ref: connect
 pre: |
-  To connect to MongoDB through the :binary:`~bin.mongo.exe <mongo>` shell,
+  To connect to MongoDB through the :binary:`~bin.mongo.exe` shell,
```
kay-kim: this I think renders as exe

You probably want :binary:`mongo.exe <bin.mongo.exe>`

atsansone: The `mongod.exe` did render as `exe`, so I filed a bug with @i80and and he fixed it because `mongod.exe` now shows normally. Why `mongo.exe` does not? I will check.

i80and: Scratch that. I see the problem. Working on a fix now

i80and: @atsansone There's no `mongo.exe` definition, since `mongo` doesn't have a Windows-specific page.

You'll have to do ```
:binary:`mongo.exe <bin.mongo>`
```

kay-kim:
I'm just going to merge and then fix on my end.

PR #3333, opened 2018-06-11
Remove statement from $and operator document

abdulhannanali:
This query can be constructed without the help of explicit `and` using `$in` operator which is encourage to be used at any place possible instead of the `$or`. It discourages learners from trying out different methods and potentially adopting $in operator. An equivalent query could be as follows. 

```js
db.inventory.find({
	price: { $in: [ 0.99, 1.99 ] },
	$or: [ { sale: true }, { qty: { $lt: 20 } } ]
});
```

We can instead add the above mentioned query here or a helpful statement. Not adding anything for now.


abdulhannanali:
Also I am so lucky to get PR #3333

PR #3348, opened 2018-06-20
(DOCS-10909): Added verification for Windows build.

atsansone:
@jdestefano-mongo : This adds information to verify a Windows MongoDB installer binary. [Staged](https://docs-mongodbcom-staging.corp.mongodb.com/anthonysansone/DOCSP-10909/tutorial/verify-mongodb-packages.html#verify-windows-packages)

@markbenvenuto : Your review would also be appreciated.

### DIFF
--- a/source/includes/steps-install-verify-files-windows.yaml
+++ b/source/includes/steps-install-verify-files-windows.yaml
@@ -1,0 +1,125 @@
+title: Download the Signature Check utility from Microsoft.
+stepnum: 1
+level: 4
+ref: download-sigcheck
+content: |
+  a. Visit the `Sigcheck utility page <https://docs.microsoft.com/en-us/sysinternals/downloads/sigcheck>`__.
+
+  b. Click the :guilabel:`Download Sigcheck` link.
+
+  c. Unzip ``Sigcheck.zip``.
+
+  d. Move the Sigcheck directory to an appropriate location on your 
+     Windows host.
+
+     For this tutorial, this location is 
+     ``$Env:ProgramFiles\Sigcheck``.
+---
+title: Download the MongoDB installation file.
+stepnum: 2
+level: 4
+ref: download-install-file
+content: |
+  Download the binaries from ``https://www.mongodb.org/downloads``.
+
+  .. example::
+
+     To download the ``v3.4-latest`` release for Windows using 
+     Powershell, invoke this command:
+
+     .. code-block:: powershell
+
+        Invoke-WebRequest -Uri "https://downloads.mongodb.org/win32/mongodb-win32-x86_64-2008plus-ssl-v3.4-latest-signed.msi" `
+          -OutFile "$Env:HomePath\Downloads\mongodb-win32-x86_64-2008plus-ssl-v3.4-latest-signed.msi"
+
+---
+title: Download the public signature file.
+stepnum: 3
+level: 4
+ref: download-sig-file
+content: |
+  Download the ``md5`` from ``https://www.mongodb.org/downloads``.
+
+  .. example::
+
+     To download the SHA256 signature for the ``v3.4-latest`` release 
+     for Windows using Powershell, invoke this command:
+
+     .. code-block:: powershell
+
+        Invoke-WebRequest -Uri "https://downloads.mongodb.org/win32/mongodb-win32-x86_64-2008plus-ssl-v3.4-latest-signed.msi.sha256" `
+          -OutFile "$Env:HomePath\Downloads\mongodb-win32-x86_64-2008plus-ssl-v3.4-latest-signed.msi.sha256"
+---
+title: Verify the signature of the MongoDB installer.
+stepnum: 4
+level: 4
+ref: verify-sig
+content: |
+
+  Invoke ``sigcheck``:
+
+  .. code-block:: powershell
+
+     $Env:ProgramFiles\Sigcheck\sigcheck64.exe `
+       -h $Env:HomePath\Downloads\mongodb-win32-x86_64-2008plus-ssl-v3.4-latest-signed.msi
+
+  ``sigcheck`` returns this verifcation information for the latest 
+  release of MongoDB 3.4:
+
+  .. code-block:: bat
+     :emphasize-lines: 19
+
+     Sigcheck v2.60 - File version and signature viewer
+     Copyright (C) 2004-2017 Mark Russinovich
+     Sysinternals - www.sysinternals.com
+
+     $Env:HomePath\downloads\mongodb-win32-x86_64-2008plus-ssl-v3.4-latest-signed.msi:
+             Verified:       Signed
+             Signing date:   12:34 AM 6/20/2018
+             Publisher:      MongoDB, Inc.
+             Company:        n/a
+             Description:    n/a
+             Product:        n/a
+             Prod version:   n/a
+             File version:   n/a
+             MachineType:    n/a
+             MD5:    D7866C013989AEE2FA87774EFFF884F0
+             SHA1:   E5D7D78E8FFFF9CFF3BD605C3407A55F87F4C8DD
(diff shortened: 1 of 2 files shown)


source/tutorial/verify-mongodb-packages.txt
```diff
@@ -10,34 +10,39 @@ Verify Integrity of MongoDB Packages
...
-+++++++++++
+~~~~~~~~~~~
 
 .. include:: /includes/steps/install-verify-files-sha.rst
+
+Verify Windows Packages
+-----------------------
+
+This verifies the MongoDB binary against its SHA256 key. This tutorial 
+uses the latest release of MongoDB Community Edition 3.4.
```
jdestefano-mongo: Why version 3.4 and not a more recent version?

source/includes/steps-install-verify-files-windows.yaml
```diff
@@ -0,0 +1,129 @@
+title: Download the Signature Check utility from Microsoft.
```
jdestefano-mongo: Being that this appears to be called Sigcheck, I think we should either say "Download the Sigcheck utility..." or lowercase "signature check"

source/includes/steps-install-verify-files-windows.yaml
```diff
@@ -0,0 +1,129 @@
...
+
+  b. Click the :guilabel:`Download Sigcheck` link.
+
+  c. Unzip ``Sigcheck.zip``.
+
+  d. Move the Sigcheck directory to an appropriate location on your 
+     Windows host.
+
+     For this tutorial, this location is 
+     ``$Env:ProgramFiles\Sigcheck``.
```
jdestefano-mongo: Is this a common notation? (I've never seen this before)

jdestefano-mongo: Also isn't it usually "Program Files" or "Program Files (x86)"? Does this run the risk of confusing readers?

source/includes/steps-install-verify-files-windows.yaml
```diff
@@ -0,0 +1,129 @@
...
+             MachineType:    n/a
+             MD5:    D7866C013989AEE2FA87774EFFF884F0
+             SHA1:   E5D7D78E8FFFF9CFF3BD605C3407A55F87F4C8DD
+             PESHA1: E5D7D78E8FFFF9CFF3BD605C3407A55F87F4C8DD
+             PE256:  8FE0670DF2AB74CCD33910C0AF2F000225BA2ED21330767D95E3F6DED96E6893
+             SHA256: 8FE0670DF2AB74CCD33910C0AF2F000225BA2ED21330767D95E3F6DED96E6893
+             IMP:    n/a
+
+---
+title: Compare the Signature File to the MongoDB Installer Hash.
```
jdestefano-mongo: I don't believe we usually titlecase step file names. Should we lowercase "signature file" and "installer hash." ?

source/includes/steps-install-verify-files-windows.yaml
```diff
@@ -0,0 +1,129 @@
...
+     :emphasize-lines: 1-2
+
+     8FE0670DF2AB74CCD33910C0AF2F000225BA2ED21330767D95E3F6DED96E6893
+     8FE0670DF2AB74CCD33910C0AF2F000225BA2ED21330767D95E3F6DED96E6893
+     True
+
+  The command outputs three lines:
+  
+  - An SHA256 hash that you downloaded directly from MongoDB.
+  - An SHA256 hash computed from the MongoDB binary you
```
jdestefano-mongo: Nit: Should this and the line above be "A SHA256 hash..."

source/includes/steps-install-verify-files-windows.yaml
```diff
@@ -0,0 +1,129 @@
...
+    downloaded from MongoDB.
+  - A ``True`` or ``False`` result depending if the hashes match.
+ 
+  If hashes match, the MongoDB binary is verified.
```
jdestefano-mongo: Nit: "If the hashes match,"

source/tutorial/verify-mongodb-packages.txt
atsansone: That's what was there originally and seemed as good a place as any. (Also, I think it's what's used in the Linux example.)

The results I show in the example are true; I actually tested them.

source/includes/steps-install-verify-files-windows.yaml
atsansone: Changed to SigCheck.

source/includes/steps-install-verify-files-windows.yaml
atsansone: This borrows from the [Microsoft Powershell Environment Variables](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_environment_variables?view=powershell-6) page and the [SS64 System Variables](https://ss64.com/nt/syntax-variables.html) page. As I was using Powershell, I wanted to use the variables just in case someone decided to configure their environment differently.

source/includes/steps-install-verify-files-windows.yaml
atsansone: Fixed.

source/includes/steps-install-verify-files-windows.yaml
atsansone: Depends how you pronounce it? In my head, I'm saying it "ESS-aych-ay" and not "SHAH". I'll change it, though, because it makes sense to do so here.

source/includes/steps-install-verify-files-windows.yaml
atsansone: Fixed.

atsansone:
@jdestefano-mongo : A last pass when you're ready. Should be quick. Restaged the changes, too.

source/includes/steps-install-verify-files-windows.yaml
jdestefano-mongo: OH, THAT MAKES SENSE NOW! I read it as SHAH... haha

source/tutorial/verify-mongodb-packages.txt
jdestefano-mongo: Ok. Would it make sense to put a note saying that the following examples use MongoDB 3.4 but the procedure works with later versions (or all versions, or list specific versions) of MongoDB?

source/includes/steps-install-verify-files-windows.yaml
jdestefano-mongo: Nit: According to the linked page the 'c' is lowercase, "Sigcheck"

source/includes/steps-install-verify-files-windows.yaml
jdestefano-mongo: Thanks for the info 👍

source/includes/steps-install-verify-files-windows.yaml
```diff
@@ -0,0 +1,129 @@
...
+        Invoke-WebRequest -Uri "https://downloads.mongodb.org/win32/mongodb-win32-x86_64-2008plus-ssl-v3.4-latest-signed.msi.sha256" `
+          -OutFile "$Env:HomePath\Downloads\mongodb-win32-x86_64-2008plus-ssl-v3.4-latest-signed.msi.sha256"
+---
+title: Verify the signature of the MongoDB installer.
+stepnum: 4
+level: 4
+ref: verify-sig
+content: |
+
+  Invoke ``sigcheck``:
```
jdestefano-mongo: Should we capitalize the 's' here?

jdestefano-mongo: If so apply to line 70

atsansone:
@kay-kim : RFM for `v3.6`, `v4.0` and `master` (in case there is a distinction).

PR #3366, opened 2018-07-02
(DOCS-9808): mongo --host supports connection strings

nlarew:
**Jira**: [DOCS-9808](https://jira.mongodb.org/browse/DOCS-9808)

**Staged**: [mongo --host](https://docs-mongodbcom-staging.corp.mongodb.com/nick/DOCS-9808/reference/program/mongo.html#cmdoption-mongo-host)

This update (finally) explicitly lists connection strings as a valid argument for the `mongo --host` option. The ticket notes a brief functionality regression in 3.4 (fixed in 3.4.2), however I'm not sure that we should list that as the hotfix was included in a minor version bump.

The update also explicitly calls out DNS seedlist connections strings (`mongodb+srv://...`) so we can remove some of the extraneous information about it. (redundant info with the main seedlist page).


source/includes/options-mongo.yaml
```diff
@@ -55,40 +55,44 @@ name: host
 directive: option
 args: <hostname>
 description: |
-  Specifies the name of the host machine where the
-  {{instance}} is running. If this is not specified,
-  {{program}} attempts to connect to a MongoDB process running on
-  the localhost.
+  Specifies the name of the host machine where the {{instance}} is
+  running. If this is not specified, {{program}} attempts to connect to
```
rkumar-mongo: If excluded / If unspecified

source/includes/options-mongo.yaml
```diff
@@ -55,40 +55,44 @@ name: host
...
+       <replSetName>/<hostname1>[:<port>][,<hostname2>[:<port>],...]
+
+  - A valid :ref:`mongodb connection string <mongodb-uri>`.
+
+    .. code-block:: sh
+
+       mongodb://[username:password@]<hostname1>[:port][,<hostname2>[:port]...][/[database][?options]]
 
-     mongodb+srv://server.example.com/?connectionTimeout=3000ms
+  - A :ref:`DNS seedlist connection string <connections-dns-seedlist>`.
```
rkumar-mongo: We should note this is 3.6+ only, just in case users think this has been around for longer. e.g. `.. versionadded: 3.6`

source/includes/options-mongo.yaml
```diff
@@ -55,40 +55,44 @@ name: host
...
+
+       mongodb+srv://<hostname>[:port][/[database][?options]]
+
+  .. note::
+
+     For TLS/SSL connections (``--ssl``), {{program}} verifies that the
+     hostname of the {{instance}} to which you are connecting matches
+     the CN or SAN of the {{instance}}'s ``--sslPEMKeyFile``
+     certificate. If the hostname does not match the CN/SAN, {{program}}
+     will fail to connect.
```
rkumar-mongo: /s/will fail/cannot connect to the {{instance}}.

source/includes/options-mongo.yaml
```diff
@@ -55,40 +55,44 @@ name: host
...
+  - A *replica set name* and the *hostname* of at least one member of
+    the replica set.
 
-  Example:
+    .. code-block:: none
 
-  .. code-block:: none
+       <replSetName>/<hostname1>[:<port>][,<hostname2>[:<port>],...]
+
+  - A valid :ref:`mongodb connection string <mongodb-uri>`.
```
rkumar-mongo: mongodb URI connection string

source/includes/options-mongo.yaml
nlarew: Oops this was already in there and I deleted it. Added the versionadded

nlarew:
@rkumar-mongo updated per your review

PR #3381, opened 2018-07-24
it should be boolean

pawarvijay:
https://stackoverflow.com/questions/25325142/how-to-set-authorization-in-mongodb-config-file


pawarvijay:
sorry its 'enabled' Not boolean - my mistake

PR #3385, opened 2018-08-07
Update dateToString.txt

dogabudak:
Added version change to the document


kay-kim:
Thank you so much for your pull request and apologies for the long delay.

Because of the way restructured text works, your change will cause that row to have 3 columns in a 2-column table.  Give the long delay, instead of asking if you'd like to resubmit, we'll go ahead and add the version added tag to the docs and close this pull.

Once again, apologies for the long delay. We do appreciate your time and effort in helping us improve the docs.

Regards,

Kay Kim

PR #3393, opened 2018-08-26
Remove password from command line

jpgklassen:
Passwords should not be typed out on the command line; this is a security issue. By simply using "-p" in the command with no password, the user will then be able to type their password without it showing up on the command line.


kay-kim:
Hi -- 
thank you so much for this pull request as well as apologies for the long delay.  The page has undergone some changes to use tabs and include additional information.  Part of that change also included the removal of the password from the command line.  As such, will close this pull request without merging. However, we appreciate your taking the time to improve the docs and apologies once again for the delay.

Regards,

Kay Kim

PR #3403, opened 2018-09-08
(DOCSP-2738): Updated MMAP caution and stable packages table.

atsansone:
@rkumar-mongo : This updates the warning at the bottom of [this section](https://docs-mongodbcom-staging.corp.mongodb.com/anthonysansone/DOCSP-2738/administration/production-notes.html#supported-platforms) and updates the table in [this section](https://docs-mongodbcom-staging.corp.mongodb.com/anthonysansone/DOCSP-2738/administration/production-notes.html#use-the-latest-stable-packages).  Links have been checked.

### DIFF
--- a/source/includes/fact-mmapv1-big-endian.rst
+++ b/source/includes/fact-mmapv1-big-endian.rst
@@ -1,6 +1,6 @@
 .. important::
 
-   :doc:`MMAPv1 </core/mmapv1>` is unsupported on all big-endian
-   architectures like the IBM z/Architecture (``s390x``) and the
-   PowerPC (``ppc64le``) architecture. MongoDB returns an error when starting a :binary:`~bin.mongod` using MMAPv1 on an unsupported architecture.
-
+   :doc:`MMAPv1 </core/mmapv1>` is supported on 64-bit Intel
+   architectures (``x86_64``) only. MongoDB returns an error when
+   starting a :binary:`~bin.mongod` using MMAPv1 on an unsupported
+   architecture.
--- a/source/includes/fact-mmapv1-big-endian.rst
+++ b/source/includes/fact-mmapv1-big-endian.rst
@@ -1,7 +1,6 @@
 .. important::
 
-   MMAPv1 is unsupported on big-endian and certain bi-endian
-   architectures. This includes the IBM z/Architecture (``s390x``) and
-   PowerPC (``ppc64le``) platforms. MongoDB returns an error if you
-   set MMAPv1 as the storage engine on ``s390x`` and ``ppc64le``
-   platforms.
+   :doc:`MMAPv1 </core/mmapv1>` is unsupported on all big-endian
+   architectures like the IBM z/Architecture (``s390x``) and the
+   PowerPC (``ppc64le``) architecture. MongoDB returns an error when starting a :binary:`~bin.mongod` using MMAPv1 on an unsupported architecture.
+
--- a/source/includes/fact-mmapv1-big-endian.rst
+++ b/source/includes/fact-mmapv1-big-endian.rst
@@ -1,5 +1,7 @@
 .. important::
 
-   MMAPv1 is not supported on big-endian architectures such as s390x.
-   MongoDB returns an error if you set MMAPv1 as the
-   storage engine on a big-endian system.
+   MMAPv1 is unsupported on big-endian and certain bi-endian
+   architectures. This includes the IBM z/Architecture (``s390x``) and
+   PowerPC (``ppc64le``) platforms. MongoDB returns an error if you
+   set MMAPv1 as the storage engine on ``s390x`` and ``ppc64le``
+   platforms.
--- a/source/includes/list-table-products-supported-architecture.rst
+++ b/source/includes/list-table-products-supported-architecture.rst
@@ -4,27 +4,35 @@ version of MongoDB products:
 
 .. list-table::
    :header-rows: 1
+   :widths: 40 15 15 15 15
    
    * - Product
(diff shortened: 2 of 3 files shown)


atsansone:
@kay-kim : Fixed.

rkumar-mongo:
Note that the entire mmapv1 section is only valid for master. There is `upcoming-4.2` wherein mmapv1 is toasted. Once we get through the cleanups, you'll need to 'forward-port' this ticket removing the mmapv1 references.

source/includes/list-table-products-supported-architecture.rst
```diff
@@ -4,27 +4,35 @@ version of MongoDB products:
...
 .. list-table::
    :header-rows: 1
+   :widths: 40 15 15 15 15
    
    * - Product
-     - x86_64/amd64
+     - x86_64
+     - ppc64le
      - s390x
-     - POWER8 (little endian)
```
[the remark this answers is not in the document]
atsansone: One is the chipset, the other is the os compatibility. If we say POWER8, we should switch the others to Intel/AMD, zSeries and ARM. I'd leave this be. We use ppc64le in many other places, including the filenames of the binaries.

source/includes/list-table-products-supported-architecture.rst
```diff
@@ -4,27 +4,35 @@ version of MongoDB products:
...
      - ARMv8-A
```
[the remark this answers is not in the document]
atsansone: Agreed. Fixed.

source/includes/fact-mmapv1-big-endian.rst
```diff
@@ -1,5 +1,7 @@
 .. important::
 
-   MMAPv1 is not supported on big-endian architectures such as s390x.
-   MongoDB returns an error if you set MMAPv1 as the
-   storage engine on a big-endian system.
+   MMAPv1 is unsupported on big-endian architectures like the IBM
+   z/Architecture (``s390x``) and certain bi-endian architectures like
+   the PowerPC (``ppc64le``) platform. MongoDB returns an error if you
```
rkumar-mongo: We should keep this second section generic as well. Otherwise it implies MongoDB does *not* return errors for any other big-endian architecture, even if MMAPv1 isn't supported.

"MongoDB returns an error when starting a {{mongod}} using MMAPv1 on an unsupported architecture."

source/includes/fact-mmapv1-big-endian.rst
```diff
@@ -1,5 +1,7 @@
...
+   set ``MMAPv1`` as the storage engine for either the ``s390x`` or
+   ``ppc64le`` platforms. MMAPv1 was deprecated in MongoDB 4.0.
```
rkumar-mongo: I realised that this include is also brought in to /core/mmapv1. Having this note here is kind of redundant as that page has a big deprecation notice at the top. I think it's OK to remove it. There shouldn't be any surprises w.r.t. MMAPv1 at this point.

source/includes/fact-mmapv1-big-endian.rst
```diff
@@ -1,5 +1,7 @@
 .. important::
 
-   MMAPv1 is not supported on big-endian architectures such as s390x.
-   MongoDB returns an error if you set MMAPv1 as the
-   storage engine on a big-endian system.
+   MMAPv1 is unsupported on big-endian architectures like the IBM
+   z/Architecture (``s390x``) and certain bi-endian architectures like
```
rkumar-mongo: I think this ultimately begs the question of what other bi-endian or little-endian archs we throw errors on (vs archs we simply don't test against for official support). If it's _only_ ppc64le that we throw errors for, we should just state that.

atsansone: Fixed but forgot to push.

source/includes/fact-mmapv1-big-endian.rst
atsansone: Removed.

source/includes/list-table-products-supported-architecture.rst
```diff
@@ -4,27 +4,35 @@ version of MongoDB products:
...
 
-   * - MongoDB 3.4
+   * - MongoDB Community (3.4+)
```
rkumar-mongo: [NP] this reverted to +. Change to "3.4 and later"

atsansone:
@acmorrow : Finally got you access to this repo. Can you have a look at this change? The staged versions are noted in the first comment.

source/includes/fact-mmapv1-big-endian.rst
```diff
@@ -1,5 +1,6 @@
 .. important::
 
-   MMAPv1 is not supported on big-endian architectures such as s390x.
-   MongoDB returns an error if you set MMAPv1 as the
-   storage engine on a big-endian system.
+   :doc:`MMAPv1 </core/mmapv1>` is unsupported on all big-endian
```
acmorrow: Rather than trying to state where it isn't supported, why not just state that it is only supported on Intel x86 processors?

acmorrow: For instance, you don't state here that it isn't supported on ARM64. The actual logic that determines whether mmapv1 is here:

https://github.com/mongodb/mongo/blob/666545a2c01ad724912391ca44be36981d3ed0d2/SConstruct#L1829-L1836

rkumar-mongo:
merged in by hand to avoid some merge conflicts: https://github.com/mongodb/docs/commit/b01adfcd509ca45b29b5b80bc1ecdcca980c86bb

PR #3409, opened 2018-09-14
(DOCS-10680): Update setspn option.

atsansone:
@rkumar-mongo : Just changes the option for `setspn.exe`. It will need backports to 3.4, 3.6.


rkumar-mongo:
@atsansone please create backport tickets for 3.6 and 3.4.

PR #3421, opened 2018-09-24
Missing examples label in "On this page" section

xxahmxx:
Changing header level of "Examples", to be visible in On this page section


kay-kim:
Thank you!

PR #3422, opened 2018-09-24
DOCS-11897 add total number of committed, aborted, and started transa…

iwysiu:
…ctions to serverStatus


kay-kim:
@iwysiu  -- could you do the pull against master? as it's been backported to 4.0.2 and as such, it's already out.

PR #3439, opened 2018-10-08
DOCS-11938 fix example for waitForSecondaryBeforeNoopWriteMS

iwysiu:


source/reference/parameters.txt
```diff
@@ -1397,7 +1397,7 @@ Replication Parameters
 
    .. code-block:: sh
 
-      mongod --setParameter waitForSecondaryBeforeNoopWriteMS=20
+      mongod --setParameter waitForSecondaryBeforeNoopWriteMS=2000
```
kay-kim: Could we make the description 20 milliseconds (instead of 20 secondes)  and put this back to 20?

This way, it's also consistent with the db.adminCommand example below

iwysiu: fixed

PR #3447, opened 2018-10-11
DOCS-7131 add embedded document example of document validation

iwysiu:


kay-kim:
@iwysiu  -- could we also update the example https://docs.mongodb.com/manual/reference/operator/query/jsonSchema/#schema-validation

iwysiu:
@kay-kim -- done

PR #3448, opened 2018-10-12
DOCS-11313 fix circular reference to Jumbo Chunk

iwysiu:


source/core/sharding-data-partitioning.txt
```diff
@@ -219,9 +219,10 @@ In some cases, chunks can grow beyond the :ref:`specified chunk size
 <sharding-chunk-size>` but cannot undergo a :ref:`split<sharding-chunk-split>`.
 The most common scenario is when a chunk represents a single shard key value.
 Since the chunk cannot split, it continues to grow beyond the chunk size,
-becoming a :ref:`jumbo chunk<jumbo-chunk>`. These jumbo chunks can become a
-performance bottleneck as they continue to grow, especially if the shard key
-value occurs with high :ref:`frequency<shard-key-frequency>`.
+becoming an indivisible chunk. These indivisible chunks, also known as jumbo
```
kay-kim: Not quite right.

So, this sentence statesthat "Since the chunk cannot split (i.e. it's indivisible), it continues to grow beyond chunk size, becoming an indivisible chunk.  Whether it continues to grow or not, it's still indivisible.

iwysiu:
fixed

PR #3460, opened 2018-10-30
Update where.txt

jtassin:
Correction on the collection name.

The doc said :

> Consider the following documents in the **users** collection

and the command was :

```javascript
db.foo.find( { $where: function() {
   return (hex_md5(this.name) == "9b53e667f30cd329dca1ec9e6a83e994")
} } );
```

This pull request just change the foo of the command to `users` according to the previous sentence

### DIFF
--- a/source/reference/operator/query/where.txt
+++ b/source/reference/operator/query/where.txt
@@ -99,7 +99,7 @@ MD5 hash and returns any matching document.
 
 .. code-block:: sh
 
-   db.foo.find( { $where: function() { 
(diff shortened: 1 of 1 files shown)


kay-kim:
Thank you so much for your pull request and apologies for the long delay.

Since the time of your pull request, we have updated the collection name through another commit, and as such, will close this pull request.

We very much appreciate your time and effort in helping us improve the docs.  Once again, apologies for the long delay. 

Regards,

Kay Kim

PR #3490, opened 2018-11-14
DOCSP-3449: Compact Oplog after Reducing Size

rkumar-mongo:


rkumar-mongo:
Backport ticket, see #3489 for root ticket

PR #3528, opened 2018-12-28
Fixed mongodump `--archive` option arguments

gfranco93:


kay-kim:
Hey @gfranco93  --
thanks for this pull and sorry for the long delay. We've actually moved away from using options files and instead just include them in the various binary files.  
And starting in 4.2, we standardized the = for all the option arguments.

https://github.com/mongodb/docs/blob/v4.2/source/reference/program/mongodump.txt#L762
https://github.com/mongodb/docs/blob/v4.2/source/reference/program/mongodump.txt#L975

As such, I'm going to go ahead and close this pull.  But lovely as always to get contributions from you.

Regards,

Kay Kim

PR #3529, opened 2019-01-09
v4.0.6

kay-kim:


kay-kim:
closing since because of conflict, manually rebased and merged

PR #3538, opened 2019-02-11
Misleading statement removed.

asya999:
All stages allow this.


asya999:
meant for this to go into master.

PR #3539, opened 2019-02-16
Fixed wrong config servers port

PaulAnnekov:
They were partially fixed here https://jira.mongodb.org/browse/DOCS-10417, but there are still occurrences of wrong ports. Fixed them.


kay-kim:
Thank you.

PR #3541, opened 2019-02-22
Update in.txt

jameswahlin:
Modified because this example match both string fields and string within an array.

### DIFF
--- a/source/reference/operator/query/in.txt
+++ b/source/reference/operator/query/in.txt
@@ -92,7 +92,7 @@ Consider the following example:
    db.inventory.find( { tags: { $in: [ /^be/, /^st/ ] } } )
 
 This query selects all documents in the ``inventory`` collection where
-the ``tags`` field holds an array that contains at least one element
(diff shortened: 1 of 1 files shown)


source/reference/operator/query/in.txt
```diff
@@ -92,7 +92,7 @@ Consider the following example:
    db.inventory.find( { tags: { $in: [ /^be/, /^st/ ] } } )
 
 This query selects all documents in the ``inventory`` collection where
-the ``tags`` field holds an array that contains at least one element
+the ``tags`` field holds a string that contains at least one element
```
kay-kim: @jameswahlin  -- could we tweak it to either a string that starts with either ``be`` or ``st`` or an array with at least one element that starts with ...

Otherwise, it sounds weird to say that a string contains at least one element ....

jameswahlin: Oh sorry, I should have read closer. Will tweak.

kay-kim: thanks mister!

PR #3543, opened 2019-02-27
Clarify that any constant can be used to group as a whole

jmikola:
Related to a discussion earlier in drivers Slack. I've left the example section further down in the document, which uses `null`, as-is; however, it'd be good to clarify that there is nothing special about `null` here and it is rather the use of a constant value that allows `$group` to utilize a single accumulation bucket.


jmikola:
Ah, I think @asya999 wins: #3542

PR #3548, opened 2019-03-18
updated UNIQUENESS AND THE _ID INDEX explanation.

manivannan89:


kay-kim:
Hi @manivannan89  -- thank you for this pull request and apologies for the delay in response.  

I believe the second bullet point is correct as is.  If you prefer to make edits to this pull, feel free.   Or we can close this pull and just update the or to a nor in the first bullet.  

Regards,

Kay Kim

PR #3561, opened 2019-04-11
DOCSP-4567: tweaks to security IA to improve organization

schmalliso:
If we decide to backport to v4.0, we'll obviously need to tweak the redirect rule accordingly.


schmalliso:
Updated!

PR #3562, opened 2019-04-14
Update transactions-production-consideration.txt

marchpig:
The content is related to `Transactions and Locks` but is duplicated in `WiredTiger Cache`.


kay-kim:
Thank you for your pull request.

PR #3570, opened 2019-05-02
3.6.13 release

kay-kim:


kay-kim:
Merged manually because of previous validate commit on v3.6.

PR #3573, opened 2019-05-09
docs($addToSet): use a simple example update for `Field is Not an Array`

gianpaj:
https://docs.mongodb.com/manual/reference/operator/update/addToSet/#field-is-not-an-array


kay-kim:
Hey @gianpaj  -- I think you meant to change a different section on the page. You seem to have updated the Value to Add is An Array example.   

Can make the change to the docs in case you don't feel like updating this.

gianpaj:
Hey @kay-kim !
Good catch. I've pushed a new commit

kay-kim:
Thank you!  merged.

PR #3574, opened 2019-05-14
DOCS-12697: Define what views can be used for

rkumar-mongo:
@kay-kim small tidbit. User feedback was that we don't really define what  view is useful for. While views are a general database concept, I felt this was an easy win for users who are new to databases and are using Mongo as their stepping stone.

### DIFF
--- a/source/core/views.txt
+++ b/source/core/views.txt
@@ -14,9 +14,23 @@ Views
    :depth: 1
    :class: singlecol
 
+MongoDB supports creating read-only views from existing collections
+or other views. For example, you can:
 
-Starting in version 3.4, MongoDB adds support for creating read-only
-views from existing collections or other views.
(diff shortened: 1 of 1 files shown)


source/core/views.txt
```diff
@@ -14,9 +14,26 @@ Views
    :depth: 1
    :class: singlecol
 
+MongoDB supports creating read-only views from existing collections
+or other views. For example, you can:
 
-Starting in version 3.4, MongoDB adds support for creating read-only
-views from existing collections or other views.
+- Create a view that :pipeline:`filters <$match>` only low-quantity
+  items from a collection of inventory data.
```
kay-kim: I don't quite know why you would need a view for this.  Could you explain a bit regarding your thinking?

rkumar-mongo: Discussed Offline, removing.

PR #3578, opened 2019-05-30
Delete duplicate paragraph about $abs and null/NaN

ksuarz:
Hi Docs Team!

The `$abs` page has a duplicate paragraph that discusses the result if the input is `null` or `NaN`, one of which comes from the extract and one of which is verbatim in the text. This pull request proposes deleting the text in the `$abs` page and keeps the extract.

Hope you're all well!
Kyle


ksuarz:
Uh oh, looks like this new Evergreen GitHub integration isn't happy.

kay-kim:
Kyle!!!!
Thank you.  Have no idea what that evergreen github stuff is :D

PR #3584, opened 2019-06-11
Added advantage to horizontal scalability

tishakariya:
Hi,

Have added an additional feature of Horizontal scaling + replica set.
Hope it adds to enhancing the MongoDB documentation.

Thanks and Regards,
Priyanka Kariya

### DIFF
--- a/source/introduction.txt
+++ b/source/introduction.txt
@@ -113,6 +113,8 @@ functionality:
   only to those shards inside the zone. See the :ref:`zone-sharding`
   manual page for more information.
 
+- Each shard in itself could have its replica set, which inturn ensures
+  high availability.
 
 Support for Multiple Storage Engines
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~


source/introduction.txt
```diff
@@ -113,6 +113,8 @@ functionality:
   only to those shards inside the zone. See the :ref:`zone-sharding`
   manual page for more information.
 
+- Each shard in itself could have its replica set, which inturn ensures
```
rkumar-mongo: As of [3.6](https://docs.mongodb.com/manual/release-notes/3.6/#sharded-clusters) shards and CSRS *must* be a replica set. 

I would also caution against saying that replication *ensures* high availability. Replication provides redundancy and high availability, but 'ensures' implies a higher guarantee level. For example, a user deploying a single-member replica set as a shard is technically using a replica set, but has no redundancy or availability guarantees. 

To summarize, three suggestions:

* Move this bullet up to be right before we talk about zoned sharding.
* Frame replication as a vehicle for redundancy and availability, but not a guarantee of it
* Sharding requires replica sets for the shard and configuration servers.

PR #3591, opened 2019-06-20
DOCS-12816: Wildcard Index Touch Ups

rkumar-mongo:


kay-kim:
@rkumar-mongo   is that extract used anywhere else?

rkumar-mongo:
Just the wildcard index page itself. I think the change actually benefits that page as well.

source/includes/extracts-wildcard-indexes.yaml
```diff
@@ -45,15 +45,21 @@ content: |
      { "userMetadata" : "inactive" }
 
   Administrators want to create indexes to support queries on any
-  subfield of ``userMetadata``. A wildcard index on ``userMetadata``
+  subfield of ``userMetadata``.
```
kay-kim: Above -- line 42 is not valid

rkumar-mongo: Fixed

PR #3594, opened 2019-06-29
Fix typo

dandv:
This bit seems out of place:

> to merge the ``pets`` document into some default document


kay-kim:
Ah - thank you.  Will fix that pets blurb.  Forgot to update after copy/paste.

PR #3608, opened 2019-07-31
Fix language driver name

alterphp:


p-mongo:
Note that:

- https://docs.mongodb.com/manual/release-notes/4.0/#drivers uses "Python"
- https://docs.mongodb.com/manual/release-notes/4.2/#drivers uses "Python"
- https://docs.mongodb.com/manual/core/security-client-side-encryption/#field-level-encryption-drivers uses "Python (PyMongo)"

If bare "Python" is the preferred spelling, this PR appears to be correct for the PHP driver. If "Python (PyMongo)" is the preferred spelling, the 4.0 and 4.2 driver listings should be changed to use that and this PR should use "PHP (PHPC)" to match.

jmikola:
> If "Python (PyMongo)" is the preferred spelling, the 4.0 and 4.2 driver listings should be changed to use that and this PR should use "PHP (PHPC)" to match.

I concur. Since this PR only pertains to PHP, I'd suggest amending the commit to "PHP (PHPC)" before merging for consistency. @p-mongo: feel free to make a separate change for Python to update any inconsistent instances in the manual, or open a DOCSP ticket to request the change.

cc: @alcaeus

alterphp:
done !

kay-kim:
Thank you!  Will merge.

PR #3616, opened 2019-08-06
Added important tag with snippet on gridfs not supporting transactions

mongomoe:
Jira ticket: https://jira.mongodb.org/browse/DOCSP-6255?filter=-1
Staging: https://docs-mongodbcom-staging.corp.mongodb.com/docsworker/DOCSP-6255-gridFsDoesntSupportTransactions/core/gridfs.html

### DIFF
--- a/source/core/gridfs.txt
+++ b/source/core/gridfs.txt
@@ -16,6 +16,13 @@ GridFS
 that exceed the :term:`BSON`\-document :ref:`size limit
 <limit-bson-document-size>` of 16 MB.
 
+
+
+.. note::
+   
+   GridFS does not support :doc:`multi-document transactions </core/transactions>`.
+
+
 Instead of storing a file in a single document, GridFS divides the file
 into parts, or chunks [#chunk-disambiguation]_, and stores each chunk as
 a separate document. By default, GridFS uses a default chunk size of 255 kB;


source/core/gridfs.txt
```diff
@@ -16,6 +16,16 @@ GridFS
...
 <limit-bson-document-size>` of 16 MB.
 
+
+
+.. important::
+
+   GridFS does not support multi-document transactions. Transactions are represented in a single oplog 
+   entry and must be under the 16MB limit. Also, transactions default lifetime is 60 seconds. Large files, 
+   like the ones you may be using GridFs, often take too long to upload. Since, GridFS supports large files 
+   above 16 MB this makes it impractical to support multi-document transactions.
```
nlarew: In 4.2 each operation within a transaction gets its own oplog entry so the 16MB limit is no longer an issue and the txn lifetime is adjustable so I'm not sure exactly where that leaves the surface area of the limitation.

I think it's probably okay to not include all of the explanation, so let's just make this a really simple note:

```rst
.. note::
   
   GridFS does not support :doc:`multi-document transactions </core/transactions>`.
```

mongomoe: Done

PR #3619, opened 2019-08-09
Update macOS bug warning

milkie:
The bug was fixed in 10.14.1, so I changed "10.13+" to specify just the 10.12 and 10.13 versions.  10.14.0 is also affected but I think we assume that users must keep up with operating system minor upgrades.


kay-kim:
Thanks Eric!

PR #3623, opened 2019-08-22
CSFLE Json Schema - added missing quote

ccho-mongodb:
UUID equivalent of a base64-encoded id in the JSON Schema was missing closing quote


rkumar-mongo:
Thanks for catching this!

PR #3627, opened 2019-08-30
(Typo fix) $ne (aggregation) refering lte comparison

nidin:
$ne (aggregation) is refering to /includes/extracts/fact-agg-comparison-expression-lte.rst
instead of fact-agg-comparison-expression-ne.rst


kay-kim:
Thank you for your fix.  This has been merged.

Regards,

Kay Kim

PR #3634, opened 2019-09-07
Update documents

slowdown2014:
Typo    read toctree

Changed to  toptree


rkumar-mongo:
@slowdown2014 `toctree` is a [Sphinx directive](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-toctree) for building a table of contents. `doctree` is not valid syntax, to my knowledge.

Was there an error encountered while building the docs that making the suggested change fixed?

PR #3635, opened 2019-09-09
Added details about build output location.

niccottrell:


README.rst
```diff
@@ -12,6 +12,8 @@ the documentation locally if you have `giza
      cd docs/
      make html
 
+If successful, the built HTML files will be located in ``build/master/html/``.
```
i80and: This isn't necessarily true; when we branch off stable releases, the build will happen under that branch's directory.

niccottrell: Updated. Is this getting too messy now?

rkumar-mongo:
@i80and looking at it, is this guide still true? there are no references to sphinx or mut here, as I'm looking through. I've got no objections to the fixups as they are otherwise.

i80and:
@rkumar-mongo Our repository READMEs are all comically out of date and shouldn't be used. But this fixup is reasonable in isolation.

PR #3672, opened 2019-09-27
added pre v4.0 redirects to atomocity page to avoid 404s

andf-mongodb:
Added pre-4.0 redirects to atomicity page for non-existent pages.

Will require backports to 3.6, 3.4, and 3.2.


kay-kim:
we don't backport redirects as we only build redirects off the master branch.

PR #3693, opened 2019-10-14
DOCSP-7259: Installation Improvement Prereq (Amazon Community)

rkumar-mongo:


source/includes/steps-install-mongodb-on-amazon-tarball.yaml
```diff
@@ -0,0 +1,60 @@
+title: Download the MongoDB ``.tgz`` tarball.
```
rkumar-mongo: General thought - these steps *seem* pretty generic, on pass 2 we might single-source these steps in particular.

source/tutorial/install-mongodb-on-amazon-tarball.txt
```diff
@@ -0,0 +1,66 @@
...
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 1
+   :class: singlecol
+
+Overview
+--------
+
+.. _`MongoDB Download Center`: https://www.mongodb.com/download-center/community?jmp=docs
```
rkumar-mongo: I wonder if perhaps we should toss this in source constants

PR #3731, opened 2019-11-06
DOCS-13090: Revise outdated production notes guidance on clock sync

rkumar-mongo:


rkumar-mongo:
Note: Will be backporting this to 3.6 as is, then modifying copy for 3.4 and 3.2.

See https://jira.mongodb.org/browse/DOCS-13090 for details on the ticket.

PR #3788, opened 2019-11-25
DOCS-9787 clarify big endian storage for ObjectId

andf-mongodb:


source/includes/fact-ObjectId-construct.rst
```diff
@@ -1,5 +1,10 @@
...
+  measured in seconds since the Unix epoch
 
-- a 5-byte random value, and
+- a 5-byte *random value*
 
-- a 3-byte counter, starting with a random value.
+- a 3-byte *incrementing counter*, initialized to a random value
+
+While the BSON format itself is little-endian, the *timestamp* and
+*counter* fields are big-endian, with the most significant bytes
```
rkumar-mongo: Sorry to dump this here, but I'd suggest changing 'fields' to 'values' - 'fields' implies there are actual fields to grab in the ObjectId JSON, where its actually just addressing that specific chunik of the byte string.

You can make this change on your local branches and force-push to your origin to update the PRs.

source/includes/fact-ObjectId-construct.rst
```diff
@@ -1,5 +1,10 @@
...
+- a 4-byte *timestamp value*, representing the ObjectId's creation,
+  measured in seconds since the Unix epoch
 
-- a 5-byte random value, and
+- a 5-byte *random value*
 
-- a 3-byte counter, starting with a random value.
+- a 3-byte *incrementing counter*, initialized to a random value
+
+While the BSON format itself is little-endian, the *timestamp* and
```
rkumar-mongo: Also, I'm not sure what 'BSON format' means - I think we're saying that the overall 12-byte string is little-endian, though two of its component parts are big-endian. I'm also unsure whether readers will be able to understand how those two can co-exist (e.g. how do you have little-endian ordering of the byte string when its component parts are big-endian?) 

Maybe we don't need to tackle that now, but I imagine in the future the question will be raised as to how exactly this plays out.

PR #3825, opened 2019-12-09
Update steps-install-mongodb-enterprise-on-red-hat.yaml

andf-mongodb:
v3.4 doesn't support the use of {+package-branch+} -- hardcoding instead.

### DIFF
--- a/source/includes/steps-install-mongodb-enterprise-on-red-hat.yaml
+++ b/source/includes/steps-install-mongodb-enterprise-on-red-hat.yaml
@@ -26,7 +26,7 @@ post: |
     {{distro_link}}. Downloads are organized by {{distro_name_full}}
     version (e.g. ``{{distro_num}}``), then MongoDB
     :doc:`release version </reference/versioning>`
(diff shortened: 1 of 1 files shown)


source/includes/steps-install-mongodb-enterprise-on-red-hat.yaml
```diff
@@ -26,7 +26,7 @@ post: |
     {{distro_link}}. Downloads are organized by {{distro_name_full}}
     version (e.g. ``{{distro_num}}``), then MongoDB
     :doc:`release version </reference/versioning>`
-    (e.g. ``{+package-branch+}``), then architecture (e.g. ``x86_64``).
+    (e.g. ``3.4``), then architecture (e.g. ``x86_64``).
```
kay-kim: just add the variable to source_constants

https://github.com/mongodb/docs/blob/v3.4/conf.py#L102

andf-mongodb: Done, with patch-12

PR #3826, opened 2019-12-09
Update conf.py

andf-mongodb:
Now supporting 'package-branch' source constant


conf.py
```diff
@@ -100,6 +100,7 @@ def has(self, *args):
     'atlas' : ('https://docs.atlas.mongodb.com%s','')
 }
 source_constants = {
+    'package-branch': '3.4',
```
kay-kim: I think using the variable **version**  should do it.  That way we only hard code one place.

PR #3830, opened 2019-12-10
Typos

ehershey:
Wording could maybe be improved even more too.. 
Something like "retrieves an object which includes attribute values corresponding to DNs this entity is a member of" or "retrieves object(s) which include attribute values corresponding to DNs this entity is a member of" 

Basically it's still not entirely clear what exactly the query is expected to retrieve but it's also not obvious to me how exactly to make it clearer.


kay-kim:
Thanks Ernie!

ehershey:
No problem! Thanks for great docs and for the quick merge!

PR #3839, opened 2019-12-18
Use "range-based sharding" spelling

p-mongo:
I think "ranged-based sharding" is not proper English


kay-kim:
Hi @p-mongo  -- thanks for this.  I'll merge this in, but just FYI, the  files in the locale/ directories are actually generated from the src/ files.  Once I merge in this, I'll regenerate (so it'll somewhat clobber your changes but still have your corrections since you corrected them in the src/ files).

PR #3857, opened 2020-01-13
Added sytemctl enable command

smitpatelx:
These commands will auto-start mongod server on each reboot and will keep mongod running. This is very important for every user, why this is not already there in the documentation?


andf-mongodb:
Hi, we have added systemd support to our install docs. You can take a look here: https://docs.mongodb.com/manual/tutorial/install-mongodb-on-ubuntu/

PR #3886, opened 2020-01-28
Kafka reference in Restrictions-on-Field-Names

alexbevi:
Field names that contain periods will fail when processed by the Kafka connector with "Invalid BSON field name". Even though this is failing due to JAVA-2810 it's worth noting explicitly in the documentation.

### DIFF
--- a/source/includes/fact-document-field-name-restrictions.rst
+++ b/source/includes/fact-document-field-name-restrictions.rst
@@ -14,4 +14,5 @@
 
      Until support is added in the query language, the use of ``$`` and
      ``.`` in field names is not recommended and is not supported by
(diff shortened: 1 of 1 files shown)


kay-kim:
Hi @alexbevi  -- thanks for this. 
Did you mean to put it here or in the kafka docs or possibly both.
Am including @jdestefano-mongo   as he  handles the kafka docs. 

Small minor note -- the changes are not quite restructured text -- so, once we know where the preferred location(s) should be, we can just close this pull and make the changes ourselves or if you prefer to edit your pull, it'll just be 

> \`MongoDB Kafka Connector <docs.mongodb.com/kafka-connector/current/>\`

PR #3963, opened 2020-02-26
add authorizedCollections parameter to the listCollections signature

p-mongo:
The parameter is documented below but was not present in the signature.


kay-kim:
Thanks Oleg!

PR #4032, opened 2020-04-01
DOCS-12492

niccottrell:
Here's my first go at a change for this DOCS ticket. I'm not sure how to test it.

### DIFF
--- a/source/reference/command/cleanupOrphaned.txt
+++ b/source/reference/command/cleanupOrphaned.txt
@@ -127,6 +127,19 @@ you may prefer to include a pause in-between iterations.
 Alternatively, to mitigate the impact of :dbcommand:`cleanupOrphaned`,
 you may prefer to run the command at off peak hours.
 
+It is also possible to tune the performance impact of :dbcommand:`cleanupOrphaned`
+with two server parameters :parameter:`rangeDeleterBatchSize`
+and :parameter:`rangeDeleterBatchDelayMS`
+
+To limit the number of documents deleted per batch, you can set
+`rangeDeleterBatchSize` to a small value such as `32`.
+
+To add an additional delay between batch deletions, you can set
+`rangeDeleterBatchDelayMS` above the current default of `20` milliseconds.
+
+Note that these affect all range deletion including normal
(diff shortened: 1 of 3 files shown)


source/reference/command/cleanupOrphaned.txt
```diff
@@ -298,4 +298,19 @@ the following:
      nextKey = result.stoppedAtKey;
    }
 
+
+Control performance impact
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+It is possible to tune the performance impact of :dbcommand:`cleanupOrphaned`
+with two server parameters `rangeDeleterBatchSize</reference/parameters/#param.rangeDeleterBatchSize>`
```
kay-kim, line 306: You can actually just go :parameter:\`rangeDeleterBatchSize\` and  :parameter:\`rangeDeleterBatchDelayMS\`

As @devkev  mentioned in the DOCS-12492, since these settings would also affect chunk migration, we probably should mention this fact -- in case people really just want to set them temporarily for the command - but want whatever for the chunk migration.

Also, probably the content should go with the Performance section -- 
https://github.com/mongodb/docs/pull/4032/files#diff-01c6bc4c93806419712b7070b0efe7e5L113

kay-kim: Oh -- and does git pull --rebase not work?  We tend to keep our commits 1 per fix.

kay-kim:
Thanks Nic.  What I'll probably do to avoid popping in the other extraneous commits is to
1) pull down your changes, 
2) cherry-pick the last 2 commits and squash, and 
3) tweak the note to .. note:: so that it becomes a little bit more prominent (just as a precaution in case people skip over since this would affect regular chunk migration) 
4) merge 
5) close this pull request

This will probably be later this evening.  Am in the middle of writing up some map-reduce -> agg counterpart examples and want to finish that up first.

Kay

kay-kim:
Merged: https://github.com/mongodb/docs/commit/48b4814f725c11d833bac272c2ac6e89243abd8a

Then tweaked:
https://github.com/mongodb/docs/commit/15902b22a4d0ee0b2fdc27bad3ca55e1a9e5e34e

Closing this pull.

PR #4057, opened 2020-04-13
[wip] Ulimit documentation improvements

p-mongo:
- Adds example output of `/proc/<pid>/limits`
- Provides this example earlier in the document, where retrieving limits is described
- Renames "/proc file system" to "proc file system" since that is the proper name (/proc is the mount point)
- Prose edited to improve flow in my opinion

### DIFF
--- a/source/reference/ulimit.txt
+++ b/source/reference/ulimit.txt
@@ -310,10 +310,11 @@ form:
 
    This section applies only to Linux operating systems.
 
-The ``proc`` file-system stores the per-process limits in the file located at
+The per-process limits in effect for a particular process can be obtained
+via the ``proc`` file system by examining the file located at
 ``/proc/<pid>/limits``, where ``<pid>`` is the process's :term:`PID` or
 process identifier. You can use the following ``bash`` function to return
-the content of the ``limits`` object for a process or processes with a
+the content of the ``limits`` file for a process or processes with a
 given name:
 
 .. code-block:: sh
--- a/source/reference/ulimit.txt
+++ b/source/reference/ulimit.txt
@@ -81,8 +81,8 @@ Review and Set Resource Limits
 ``ulimit``
 ~~~~~~~~~~
 
-You can use the ``ulimit`` command at the system prompt to check
-system limits, as in the following example:
+You can use the ``ulimit`` command at the system prompt to check resource
+limits applicable to the running shell, as in the following example:
 
 .. code-block:: sh
 
@@ -103,6 +103,12 @@ system limits, as in the following example:
    -e: max nice                   30
    -r: max rt priority            65
    -N 15:                         unlimited
+
+.. note::
+
+  If :binary:`~bin.mongod` or :binary:`~bin.mongos` processes are launched
+  using a systemwide process manager such as Upstart or systemd, they may
+  be subject to different resource limits.
 
 ``ulimit`` refers to the per-*user* limitations for various
 resources. Therefore, if your :binary:`~bin.mongod` instance executes as a
--- a/source/reference/ulimit.txt
+++ b/source/reference/ulimit.txt
@@ -140,15 +140,57 @@ For many distributions of Linux you can change values by substituting
 the ``-n`` option for any possible value in the output of ``ulimit
 -a``.
 
-After changing the ``ulimit`` settings, you *must* restart the
-process to take advantage of the modified settings. On Linux, you can
-use the ``proc`` file system to see the current limitations on a
-running process.
+When ``ulimit`` is used to change resource limits in a shell, the changes
+affect only the processes subsequently launched by that shell process.
+This means:
 
-Depending on your system's configuration, and default settings, any
-change to system limits made using ``ulimit`` may revert following
-a system restart. Check your distribution and operating
-system documentation for more information.
+- If you launch :binary:`~bin.mongod` or :binary:`~bin.mongos` manually from
+  a shell, you must start new :binary:`~bin.mongod` or :binary:`~bin.mongos`
+  processes after changing resource limits via ``ulimit``.
+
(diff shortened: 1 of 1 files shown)


source/reference/ulimit.txt
```diff
@@ -104,6 +104,12 @@ system limits, as in the following example:
    -r: max rt priority            65
    -N 15:                         unlimited
 
+.. note::
+
+  If :binary:`~bin.mongod` or :binary:`~bin.mongos` processes are launched
+  using a systemwide process manager such as Upstart or systemd, they may
+  be subject to different resource limits.
```
andf-mongodb, line 111: I think this note would be more helpful if it pointed out that ulimit considerations are already handled in MDB init scripts. See the init.d, service, upstart, suse files here:

https://github.com/mongodb/mongo/tree/master/debian
https://github.com/mongodb/mongo/tree/master/rpm

Even brew sets appropriate ulimit values here:

https://github.com/mongodb/homebrew-brew/blob/master/Formula/mongodb-community.rb

So instead of differentiating based on running shell vs system-started shell,  I would differentiate via user action required (TGZ install) vs no action required (package manager install).

source/reference/ulimit.txt
```diff
@@ -303,22 +351,23 @@ form:
...
 
    This section applies only to Linux operating systems.
 
-The ``/proc`` file-system stores the per-process limits in the
-file system object located at ``/proc/<pid>/limits``, where ``<pid>``
-is the process's :term:`PID` or process identifier. You can use the
-following ``bash`` function to return the content of the ``limits``
-object for a process or processes with a given name:
+The per-process limits in effect for a particular process can be obtained
+via the ``proc`` file system by examining the file located at
```
andf-mongodb, line 362: Singular "file" makes me think that I am checking one config file for all processes. I would recommend something like:

" ... can be obtained via the ``proc`` filesystem. Each running process stores its current ``ulimit`` values in a ``limits`` file located at ``/proc/<pid>/limits``, where ... "

source/reference/ulimit.txt
```diff
@@ -303,22 +351,23 @@ form:
...
+``/proc/<pid>/limits``, where ``<pid>`` is the process's :term:`PID` or
+process identifier. You can use the following ``bash`` function to return
+the content of the ``limits`` file for a process or processes with a
+given name:
 
 .. code-block:: sh
 
-   return-limits(){
+   get-limits(){
```
andf-mongodb, line 370: Good change!

source/reference/ulimit.txt
```diff
@@ -140,15 +146,57 @@ For many distributions of Linux you can change values by substituting
...
+This means:
+
+- If you launch :binary:`~bin.mongod` or :binary:`~bin.mongos` manually from
+  a shell, you must start new :binary:`~bin.mongod` or :binary:`~bin.mongos`
+  processes after changing resource limits via ``ulimit``.
+
+- If :binary:`~bin.mongod` or :binary:`~bin.mongos` are launched using
+  startup scripts, the ``ulimit`` invocation must generally be placed in those
+  startup scripts prior to :binary:`~bin.mongod` or :binary:`~bin.mongos`
+  invocations.
```
andf-mongodb, line 160: Unless users  installed MDB via the TGZ, and are creating their own init scripts (init.d, systemd, upstart), the appropriate ulimit values are already present in init scripts provided with the RPM or DEB file they installed. 

It's still valuable to discuss what to do for users with custom scripts (For example, the numactl tabs here https://docs.mongodb.com/manual/administration/production-notes/#configuring-numa-on-linux), but it is the least likely case for our users. I don't think we need anything more than a small mention. So the three options are:

1. Run from command line: you must set ulimit via the methods provided on this page.
2. Run from our init script: no action
3. Installed via TGZ and made your own init script: you must edit your init script to set ulimit

source/reference/ulimit.txt
```diff
@@ -140,15 +146,57 @@ For many distributions of Linux you can change values by substituting
...
+
+  .. note::
+
+    On Linux distributions using Upstart, systemd or a similar process
+    manager to start :binary:`~bin.mongod` or :binary:`~bin.mongos` processes,
+    the resource limits must be configured in the process manager
+    as described below.
```
andf-mongodb, line 167: Not necessary, see above.

source/reference/ulimit.txt
```diff
@@ -140,15 +146,57 @@ For many distributions of Linux you can change values by substituting
...
+  Max processes             62913                62913                processes 
+  Max open files            20000                20000                files     
+  Max locked memory         16777216             16777216             bytes     
+  Max address space         unlimited            unlimited            bytes     
+  Max file locks            unlimited            unlimited            locks     
+  Max pending signals       62913                62913                signals   
+  Max msgqueue size         819200               819200               bytes     
+  Max nice priority         0                    0                    
+  Max realtime priority     0                    0                    
+  Max realtime timeout      unlimited            unlimited            us
```
andf-mongodb, line 194: I think the bulk of this is adequately addressed below in the proc filesystem section, no? Let's link to that section from option 1 (run from command line) above. This method does not apply to method 2 or 3. I would use ``get-limits()`` for those two methods to verify limits.

If you like, the ``/proc/<PID>/limits`` output could go down there as well, but I think it largely duplicates this output earlier:

https://docs.mongodb.com/manual/reference/ulimit/#ulimit

p-mongo:
I can certainly leave this PR until the redesign, but I am having trouble understanding the following:

> I think the key consideration is that ulimit values are already set in the appropriate init scripts

Whatever values may be set in the init scripts, they may need to be adjusted up (if they are set conservatively and the particular deployment has many resources) or down (if they are set greedily and the particular deployment wants to be constrained). Therefore I would think that simply saying that init scripts that come with MongoDB set the values "appropriately" is not particularly  helpful for administrators who need to change those values to suit their environment.

PR #4067, opened 2020-04-15
Clarify the versions are for the driver library

mltsy:
Without the driver names, it looks like the version numbers might be for the language rather than the driver (which is confusing, at least for Ruby, which is currently at v2.7)

### DIFF
--- a/source/includes/3.6-drivers.rst
+++ b/source/includes/3.6-drivers.rst
@@ -1,21 +1,21 @@
 .. list-table::
    :class: index-table
 
-   * - Java 3.6+
+   * - Java: mongodb-driver-sync 3.6+
 
-       Python 3.6+
+       Python: pymongo 3.6+
 
-       C 1.9+
+       C: libmongoc 1.9+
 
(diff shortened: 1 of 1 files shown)


p-mongo:
This is an interesting idea. As an alternative, perhaps linking to each driver's documentation starting page would clarify that the version number applies to the driver and not to the language of the driver.

mltsy:
Sure, no problem with linking to the driver pages, other than someone would have to maintain those links, but I still think it should list the driver name here rather than just the language name... is there any issue with listing the driver names?  (Even without links, this makes it much easier to find the appropriate driver in a search)

source/includes/3.6-drivers.rst
```diff
@@ -1,21 +1,21 @@
 .. list-table::
    :class: index-table
 
-   * - Java 3.6+
+   * - Java: mongodb-driver-sync 3.6+
```
p-mongo, line 4: Does the async driver also needs to be mentioned here?

PR #4068, opened 2020-04-16
DOCSP-9931 clarify temporary tools download language

andf-mongodb:


source/includes/extracts-migration-to-database-tools.yaml
```diff
@@ -0,0 +1,73 @@
+ref: _migration-to-database-tools
+content: |
+  .. admonition:: Upcoming Tools Migration in MongoDB 4.4
+    :class: note
+
+    Starting in MongoDB 4.4, the ``{{command}}`` documentation 
+    migrates to `MongoDB Database Tools - {{command}} 
+    <https://docs.mongodb.com/database-tools/{{command}}>`__. The
+    latest version of ``{{command}}`` (tool version 100.0.0 and
```
kay-kim, line 9: @andf-mongodb  -- Oh, the latest could be construed to mean the current latest.  I would use the word "upcoming" since we're already using it in the title -- The upcoming version of ...

source/includes/extracts-migration-to-database-tools.yaml
```diff
@@ -0,0 +1,73 @@
...
+    later) is now available for download following the instructions on
+    that page.
```
kay-kim, line 11: Is "is now available for download ..." true as page sends me back to 4.2 manual?

I'd just be like ", refer to the  MongoDB Database Tools  documentation" or something like that.

source/includes/extracts-migration-to-database-tools.yaml
andf-mongodb: Removed this line entirely in favor of ...

PR #4088, opened 2020-04-28
correction to scope of metric

asya999:
serverStatus shows metrics for mongod as a whole, not for an individual database.


kay-kim:
thank you!

PR #4100, opened 2020-05-04
Fix typo in intentionally non-existent field

benatuts:
The paragraph above refers to an intentionally non-existent field $qty. This commit fixes the code so that it matches the non-existent field (rather than the real field).


kay-kim:
Thank you!  This has been merged.

PR #4103, opened 2020-05-09
Change single quotes to double quotes

melkadze:
Single quotes in the --nsFrom and --nsTo parameters are NOT understood through the command line mongorestore tool, and prevent it from working as intended (instead of replacing test.*, it replaces 'test.*') . Double quotes maintain the consistency of the similarly double-quoted command arguments, and allow this command to function normally.


p-mongo:
@melkadze What operating system do you use?

melkadze:
> @melkadze What operating system do you use?

@p-mongo This was on the Windows command line

PR #4105, opened 2020-05-11
Fix tls/ssl language

p-mongo:


source/tutorial/configure-x509-member-authentication.txt
```diff
@@ -82,9 +82,9 @@ Use Command-line Options (``tls``)
 
 .. note::
 
-   The procedures in this section use the ``ssl`` settings/option. For
-   procedures using their ``tls`` aliases (Available in MongoDB 4.2),
-   see :ref:`configure-member-tls`.
+   The procedures in this section use the ``tls`` settings/option. For
```
rkumar-mongo, line 85: Can we use the existing copy from https://github.com/mongodb/docs/blob/master/source/tutorial/configure-x509-client-authentication.txt#L66-L74 ?

I'm hesitant to write in 'all versions' as we may remove SSL in the future (and have to go back and remember to fix this at that point in time). 

Alternatively, tweak to just state 'For procedures using the deprecated ``ssl`` aliases, see <blah>'.

p-mongo: Changed to the last paragraph, existing wording doesn't make sense (ssl options are available before 4.2).

rkumar-mongo:
LGTM - @p can you squash down to a single commit? I'll merge and backport to 4.2 after.

p-mongo:
Github allows squash merge, wouldn't that be equivalent?

rkumar-mongo:
It's disabled for this repository (I think we've had issues getting it to work in the past without adding an ugly merge commit in our tree)

p-mongo:
Okay I squashed them.

rkumar-mongo:
Thanks @p-mongo ! I'll backport to 4.2.

PR #4127, opened 2020-06-02
fix to aggregation skip example

AlainBeckerCPE:
Sample code seems incorrect to me: aggregate() argument is an array
BEFORE : db.article.aggregate({ $skip : 5 });
AFTER : db.article.aggregate([{ $skip : 5 }]);

Willing to know if this ought to be ported to all version branches (and if not all, all from... which version ?).

### DIFF
--- a/source/reference/operator/aggregation/skip.txt
+++ b/source/reference/operator/aggregation/skip.txt
@@ -35,9 +35,9 @@ Consider the following example:
 
 .. code-block:: javascript
 
-   db.article.aggregate(
+   db.article.aggregate([
        { $skip : 5 }
-   );
+   ]);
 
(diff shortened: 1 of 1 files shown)


kay-kim:
Hi - Thank you for the pull request.  We much appreciate your taking the time to help us improve the docs.  The page was indeed using the legacy format.  In the mongo shell, the shell still supports the legacy format (see below), but definitely should be updated.  Your pull request has been merged and we'll backport to older versions.  Thanks again.
> 
> function(pipeline, aggregateOptions) {
>     if (!(pipeline instanceof Array)) {
>         // Support legacy varargs form. Also handles db.foo.aggregate().
>         pipeline = Array.from(arguments);

PR #4129, opened 2020-06-03
See also $pull for $addToSet

dandv:


kay-kim:
Thanks as always Dan.  This has been merged.

PR #4135, opened 2020-06-08
Added a 'snapshot' option to readConcernLevel in connection string

hyunjinjeong:
It seems that a `snapshot` option of `readConcernLevel` has been missed in the document.


p-mongo:
Per https://docs.mongodb.com/manual/reference/read-concern-snapshot/ the snapshot read concern can only be used in transactions. It is technically possible to set read concern to snapshot globally, but if this is done all reads must be done in transactions on the client so constructed.

If the change proposed in this PR is made, it should be accompanied by a note stating that all read operations on a client constructed with snapshot read concern must be performed in transactions.

PR #4168, opened 2020-06-23
correcting where merging happens

asya999:
Removed incorrect reference to primary shard

This should be fixed in master and 4.2 and 4.0


kay-kim:
Thanks Asya.  There's a https://jira.mongodb.org/browse/DOCS-11469 ticket that will further clean up.

PR #4184, opened 2020-07-06
Update parameters.txt about parameters reverting to default after reboot unless set in config file or cmd line opt

bslade:
Added a note saying parameters set via the :dbcommand:`setParameter` option will revert to default after reboot unless set in the config file or command line options.

This may seem obvious once you know it, but it's different from a lot of other database systems.

Ben Slade
DBA @ NCBI.NLM.NIH.gov

### DIFF
--- a/source/reference/parameters.txt
+++ b/source/reference/parameters.txt
@@ -22,7 +22,7 @@ using:
 
      db.adminCommand( { setParameter: 1, <parameter>: <value>  } )
 
-- the :setting:`setParameter` configuration setting:
+- the :setting:`setParameter` configuration setting in the boot time configuration file:
 
   .. code-block:: yaml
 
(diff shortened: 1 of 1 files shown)


source/reference/parameters.txt
```diff
@@ -22,7 +22,7 @@ using:
 
      db.adminCommand( { setParameter: 1, <parameter>: <value>  } )
 
-- the :setting:`setParameter` configuration setting:
+- the :setting:`setParameter` configuration setting in the boot time configuration file:
```
p-mongo, line 25: I think there is only ever one configuration time, hence adding "boot time" just seems confusing to me.

source/reference/parameters.txt
```diff
@@ -42,6 +42,8 @@ For additional configuration options, see
 :doc:`/reference/configuration-options`, :binary:`~bin.mongod` and
 :binary:`~bin.mongos`.
 
+Changes made via the :dbcommand:`setParameter` will revert to defaults after reboot unless they're included in either the boot time configuration file or the ``--setParameter`` command-line option.
```
p-mongo, line 45: I think this note should be under the section that deals with setParameter call for further clarity.

source/reference/parameters.txt
bslade: This was my ignorance.  I had the idea in my head that the parameters were somehow stored somewhere independently of the startup command line (like other databases I use).  I'd like to withdraw this suggested edit.

PR #4202, opened 2020-07-15
Correct typo with grep command.

metasansana:
I think this was a typo, the command works fine without the extra `grep -v grep`. Please ignore if it was not.

### DIFF
--- a/source/tutorial/install-mongodb-on-os-x.txt
+++ b/source/tutorial/install-mongodb-on-os-x.txt
@@ -184,7 +184,7 @@ assume that you are using the default settings.
 
    .. code-block:: sh
 
-      ps aux | grep -v grep | grep mongod
+      ps aux | grep mongod
 
    You can also view the log file to see the current status of your
    ``mongod`` process: ``/usr/local/var/log/mongodb/mongo.log``.


andf-mongodb:
While you'll still see the `mongod` process either way, the `ps` output will include the `grep` command itself when run without piping to `grep -v grep`, which might be confusing to some users. Shown here:

`~ $ ps aux | grep mongod
joe 39558   0.0  0.1  5581196  12588   ??  S    Wed05PM   6:10.82 mongod --dbpath /usr/local/var/mongodb --logpath /usr/local/var/log/mongodb/mongo.log --fork
joe 97095   0.0  0.0  4277296    852 s005  S+    5:25PM   0:00.00 grep mongod
~ $
~ $ ps aux | grep -v grep | grep mongod
joe 39558   0.2  0.1  5581748  12568   ??  S    Wed05PM   6:10.87 mongod --dbpath /usr/local/var/mongodb --logpath /usr/local/var/log/mongodb/mongo.log --fork
~ $`

The hope was that this extra bit of command syntax staves off potential confusion when looking at the results. Thanks for reaching out though -- feedback always welcome!

PR #4221, opened 2020-07-22
DOCSP-11308: [Server] Minor tweaks to manual landing page

steveren:
[Jira](https://jira.mongodb.org/browse/DOCSP-11308)
[Staged](https://docs-mongodbcom-staging.corp.mongodb.com/stevenrenaker/DOCSP-11308/index.html)

### DIFF
--- a/source/index.txt
+++ b/source/index.txt
@@ -33,7 +33,7 @@ database:
   features such as LDAP and Kerberos support, on-disk encryption, 
   and auditing.
 
-- `MongoDB Atlas <https://www.mongodb.com/cloud/atlas?tck=docs_server>`__
+- `MongoDB Atlas <https://www.mongodb.com/cloud/atlas?jmp=docs_server>`__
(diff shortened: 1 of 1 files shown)


source/index.txt
```diff
@@ -33,6 +33,10 @@ database:
   features such as LDAP and Kerberos support, on-disk encryption, 
   and auditing.
 
+- `MongoDB Atlas <https://www.mongodb.com/cloud/atlas?jmp=docs_server>`__
```
nathan-contino-mongo, line 36: The ticket specifies that we ought to use `tck` instead of jmp here -- maybe you know something I don't (from the comments, it looks like you might just want to cover that in a separate ticket)? :smile:

steveren: Yeh, we're going to replace it globally which is a bigger issue and for revertability I'll do it in a separate commit.

PR #4252, opened 2020-08-12
update document

webapis:
It seems to me that "Update  Values in an Array" phase means $ postional operator changes all values inside array. "Update  a Value in an Array" is more close to what $ is going todo. 
The same is about  Update Documents in an Array. It would be more close what $ positional operator is going to do if we change it to "Update a Document in an Array".

### DIFF
--- a/source/reference/operator/update/positional.txt
+++ b/source/reference/operator/update/positional.txt
@@ -97,7 +97,7 @@ operator to update this field.
 Examples
 --------
 
-Update Values in an Array
+Update a Value in an Array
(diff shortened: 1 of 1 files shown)


jeff-allen-mongo:
I'd like to think more on this a bit. My concern is that just because the examples use `updateOne()`, the more general application is that you could use either `updateMany()` or `updateOne()`, in which case the current wording would be more accurate.

I'm slightly concerned with this edit possibly leading readers to think they couldn't use this operator to affect multiple documents in the collection.

PR #4264, opened 2020-08-18
fixed typo in replication.txt: mongod -> mongodb

ZhuoxiLi:


p-mongo:
There is no problem with existing language. Technically both "mongod instance" and "mongodb instance" could be valid, but "mongod instance" is used earlier in the text to refer to the daemon binary.

andf-mongodb (commented):
Oleg, this should be {{:binary:`~bin.mongod`}}

p-mongo:
I made https://github.com/mongodb/docs/pull/4397 to fix that. @andf-mongodb

PR #4295, opened 2020-08-25
Correct note on deprecation

asya999:
Ensure index is not an alias for create index

### DIFF
--- a/source/reference/method/db.collection.ensureIndex.txt
+++ b/source/reference/method/db.collection.ensureIndex.txt
@@ -20,7 +20,7 @@ Definition
 
 
    .. deprecated:: 3.0.0
-      :method:`db.collection.ensureIndex()` is now an alias for
+      :method:`db.collection.ensureIndex()` has been replaced by 
       :method:`db.collection.createIndex()`.
 
    Creates an index on the specified field if the index does not


p-mongo:
On 4.4 ensureIndex is defined as follows:

```
MongoDB Enterprise ruby-driver-rs:PRIMARY> db.c.ensureIndex
function(keys, options) {
    var result = this.createIndex(keys, options);

    if (this.getMongo().writeMode() != "legacy") {
        return result;
    }

    err = this.getDB().getLastErrorObj();
    if (err.err) {
        return err;
    }
    // nothing returned on success
}
```

Seems like it just calls createIndex?

asya999:
It doesn’t pass all the options at least in 4.2

We are removing it in the next version anyway.

p-mongo:
4.2.8 shell shows:

```
MongoDB Enterprise mongos> db.x.ensureIndex
function(keys, options) {
    var result = this.createIndex(keys, options);

    if (this.getMongo().writeMode() != "legacy") {
        return result;
    }

    err = this.getDB().getLastErrorObj();
    if (err.err) {
        return err;
    }
    // nothing returned on success
}
```

asya999:
@p calling createIndex with three parameters has SIGNIFICANTLY different effect than calling ensureIndex with three parameters.   Alias implies the same effect.

p-mongo:
I see, that makes sense.

p-mongo (approved):

andf-mongodb:
Thank you! Merged & will backport.

PR #4298, opened 2020-08-26
Spelling correction: then to than in source/reference/operator/aggregation/bucketAuto.txt

kevinbai-cn:
Spelling correction: then to than in source/reference/operator/aggregation/bucketAuto.txt


jeff-allen-mongo:
Thanks @kevinbai-cn !

PR #4312, opened 2020-09-01
DOCS-13703 drop collection sharded environments need cache refreshed

jason-price-mongodb:


source/reference/method/db.collection.drop.txt
```diff
@@ -26,6 +26,18 @@ Definition
...
+
+      For a sharded cluster, if you use :method:`db.collection.drop()`
+      and then create a new collection with the same name, you must
+      either:
+   
+      - Flush the cached routing table on every :binary:`~bin.mongos`
+        using :dbcommand:`flushRouterConfig`.
+      - To avoid flushing the cache, you can use
+        :method:`db.collection.remove()` to remove the existing
+        documents and reuse the collection.
```
jeff-allen-mongo, line 39: Can we please make these bullets more parallel in structure prior to merge?
```suggestion
      - Use :method:`db.collection.remove()` to remove the
        existing documents and reuse the collection. Use this
        approach to avoid flushing the cache.
```

jason-price-mongodb: Done per your recommendation.

PR #4378, opened 2020-09-23
DOCSP-12336 Operator query in

jason-price-mongodb:


jeff-allen-mongo:
Just a note: For future PRs, when you're using the same content in multiple places, consider creating an `include` for easier future maintainability. Not a huge deal here since this is a small change, but something to keep in mind.

PR #4383, opened 2020-09-24
DOCS-13593 Add seed list definition and details

davemungo:


source/reference/connection-string.txt
```diff
@@ -202,18 +202,26 @@ DNS Seedlist Connection Format
...
+DNS SRV record. The driver or :binary:`~bin.mongo` shell will then
+query the DNS for the record to determine which hosts are running the
+:binary:`~bin.mongod` instances.
 
-The ``+srv`` indicates to the mongo client that the hostname that follows corresponds to a 
-DNS SRV record. The client driver will then query the DNS for the record to determine the hosts that
-are running the mongod instances.
+.. note::
+   Use of the +srv connection string modifier automatically sets the
+   tls (or the equivalent ssl) option to true for the connection. You
```
jeff-allen-mongo, line 218: Hey Dave, sorry to request edits at this stage, but for tls and ssl, can we please use the `:urioption:` as done in [the master branch version](https://github.com/mongodb/docs/blob/master/source/reference/connection-string.txt#L229) of this file?

I think it's important to link out to those options and they'll get formatted nicely with monospace text.

jeff-allen-mongo: Oh wait, actually `tls` is new in v4.2. I don't think we want this note on v3.6.

jeff-allen-mongo: Ah, I see that the tls option is new in v4.2, so we can't link to it directly from in v3.6. I would suggest either using the double-backtick around `tls` and `ssl` to have them render as monospace, or reword such that we don't mention tls on versions where it's not available.

cc @andf-mongodb

davemungo:
No problem, not sure how it got deleted in the first place

Dave

jeff-allen-mongo:
Thanks @davemungo, please make sure you see my second comment for why `urioption` won't work here in case you missed it 😉

PR #4631, opened 2020-12-04
remove reference to natural

asya999:
Natural sort order is unstable so shouldn't be used in an example


andf-mongodb:
Thanks @asya999 !

PR #4727, opened 2020-12-29
Update manage-indexes.txt

lesterw1:
Include information on how to force MongoDB to use a *particular* index. Examples were copied from https://docs.mongodb.com/manual/tutorial/measure-index-use/.


lesterw1:
It may be necessary to update the "On this Page" section at the top to include a reference to the **Using a particular Index** section within.

PR #4733, opened 2020-12-30
Update setIntersection.txt

Rishi174:
the second example of $setintersection should have resulted ["a","b"]. Instead in the docs it was shown as to be resulting an empty array


Rishi174:
My bad. didn't look at it properly

PR #4741, opened 2021-01-04
DOCS-13985 clarifies when compound indexes can contain a hashed index…

jmd-mongo:
… field


jeff-allen-mongo:
Going to close this out since I don't think we need to backport this back to 4.2.

PR #4750, opened 2021-01-05
DOCS-14134: remove incorrect statement about no query

asya999:
No query is allowed on *the projected* array, not on no array.  See SERVER-53576 (aka DOCS-14134)


asya999:
remove incorrect statement

andf-mongodb:
Thank you @asya999 . Resolved in DOCS-14134.

PR #4754, opened 2021-01-06
DOCS-14225: Remove incorrect TTL reference

asya999:
TTL uses replication not clock sync on members of a replica set.

### DIFF
--- a/source/administration/production-notes.txt
+++ b/source/administration/production-notes.txt
@@ -643,14 +643,6 @@ operations like the following:
 - Features which rely on timekeeping may have inconsistent or
   unpredictable behavior in clusters with clock drift between MongoDB
   components. 
-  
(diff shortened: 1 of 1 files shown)


asya999:
Note that originally this paragraph was supposed to describe two *shards* in a cluster being out of sync "time-wise" - and some cases where document gets migrated and deleted earlier than it would have in the originally shard, but the text added instead implies that replica set members need to be in sync for TTL to work, which isn't right (and TTL collection doc correctly describes that secondaries delete documents via replication mechanism and TTL thread only runs on primaries.

andf-mongodb:
Thanks @asya999 ! I agree with your helpful comment, and have removed the example paragraph. Will port forward & back as needed. Thank you!

PR #4773, opened 2021-01-11
DOCS-13883 add  random number generator

davemungo:


source/includes/extracts-agg-operators.yaml
```diff
@@ -489,6 +489,11 @@ content: |
       * - Name
         - Description
 
+      * - :expression:`$rand`
+        - Returns a random float between 0 and 1
+        
+          .. versionadded:: 4.4.2
+
```
andf-mongodb, line 496: Can you add this same `versionadded` to the `$sampleRate` entry just beneath as well please? Looks like I missed this in the review for that backport.

source/reference/operator/aggregation.txt
```diff
@@ -647,7 +647,13 @@ Alphabetical Listing of Expression Operators
...
 
      - Converts a value from radians to degrees.
-   
+
+
+   * - :expression:`$rand`
+
+     - Returns a random float between 0 and 1.
+
+       .. versionadded:: 4.4.2
```
andf-mongodb, line 656: Same here: please add `versionadded` to `$sampleRate`, below.

andf-mongodb (changes requested):
Tiny addition please: let's add the same `versionadded` to the two places where `$sampleRate` is listed. Thanks!

andf-mongodb (approved):

PR #4866, opened 2021-02-02
update link and description to drivers landing page

ccho-mongodb:
In this PR, I updated the description of the drivers section to reflect its current purpose.
One question is whether you'd prefer to keep the link to "/ecosystem" which currently redirects to "/drivers" for analytics purposes or whether it is ok to update to the latter.


terakilobyte (approved):
lgtm

PR #4940, opened 2021-02-17
DOCSP-7693 resumable index build

davemungo:


source/reference/command/createIndexes.txt
```diff
@@ -97,6 +97,10 @@ Definition
           marks the ``indexes`` as ready. A "voting" member is any
           replica set member where :rsconf:`members[n].votes` is greater
           than ``0``.
+
+          Starting in MongoDB v5.0 it is possible to resume some
+          :ref:`interupted index builds<index-operations-build-failure>`
+          when the commit quorum is set to ``"votingMembers"``.
```
jeff-allen-mongo, line 103: Two notes here:

1. Can we please add a link to instructions on how to set the commit quorum as done elsewhere where you linked to the `createIndexes-cmd-commitQuorum` ref?
2. Let's use second person. "Starting in MongoDB 5.0, you can resume some..."

PR #5130, opened 2021-03-25
DOCS-13943 Add plan cache limit and stats

npentrel:


source/reference/operator/aggregation/planCacheStats.txt
```diff
@@ -175,24 +180,36 @@ including:
...
 
      - A boolean that indicates whether the an :ref:`index filter
        <index-filters>` exists for the query shape.
 
+   * - ``estimatedSizeBytes``
+
+     - A number that describes the estimated size in bytes of a plan
+       cache entry.
+
+       .. versionadded:: 4.2.12 (and 4.0.23, and 3.6.23*)
```
jeff-allen-mongo, line 210: Looks like there might be a stray asterisk after 3.6.23?

npentrel: Oh there's actually a missing one... I'll fix it :)

PR #5163, opened 2021-04-07
Changed log message ID type from String to Integer

corentinaltepe:
I noticed an error in the log message documentation with mongo v4.4.
I was receiving log message with an `"id"` being an `Integer`, and not a `String`. Example:

```json
{
  ...
  "id": 123456,
  ...
}
```

So I offer to update documentation.

### Porting to other branches?
I have not tested any other version of mongo, nor compared with the doc for other branches. I would expect this change should be ported to other branches but I couldn't evaluate which ones.

Related work where the inconsistency was noticed: https://github.com/Mongo2Go/Mongo2Go/pull/114

### DIFF
--- a/source/reference/log-messages.txt
+++ b/source/reference/log-messages.txt
@@ -98,7 +98,7 @@ specification, and has the following layout and field order:
      "s": <String>, // severity
      "c": <String>, // component
      "ctx": <String>, // context
-     "id": <String>, // unique identifier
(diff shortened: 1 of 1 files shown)


andf-mongodb:
Thank you, you are correct. I've merged this in to primary v4.4 branch via https://github.com/mongodb/docs/pull/5336

PR #5178, opened 2021-04-08
DOCS-13943 Add plan cache limit and stats

npentrel:


source/core/query-plans.txt
```diff
@@ -72,6 +72,29 @@ The plan cache does not persist if a :binary:`~bin.mongod` restarts or shuts dow
    entire plan cache. Users can also clear particular plan cache entries using
    :method:`PlanCache.clearPlansByQuery()`.
 
+Plan Cache Debug Info Size Limit
+--------------------------------
+
+Starting in 4.0.23 (and 3.6.23), the
```
jeff-allen-mongo, line 78: Should this just say 3.6.23 since we are in the 3.6 docs? I'm not sure if we want to mention 4.0.X on the 3.6 docs.

PR #5217, opened 2021-04-20
DOCSP-15632: add C driver to CSFLE compatibility table

terakilobyte:
I came across this during triage.


andf-mongodb:
Thanks Nathan! I addressed this outside of this PR, and fore/backported. Thanks for this helpful baseline!

PR #5285, opened 2021-05-03
DOCSP-15852 remove obsolte pages db.eval db.getProfilingLevel

davemungo:
This update is part of a larger effort to remove obsolete references to unsupported methods in mongosh. 

[DOCSP-15852](https://jira.mongodb.org/browse/DOCSP-15852) removes pages db.eval() and db.getProfilingLevel() since they were deprecated a few releases ago. 

The only copy updates are to delete the pages, remove them from the nav indexes, and to convert former links to monospace. 

The removed pages have also been added to the redirect list.


addaleax (approved):
Looks good, but fwiw, there's quite a few more db methods that mongosh does not currently support.

jeff-allen-mongo:
The redirects for this PR were added in 4b37658f51b3785f3780a1dc5b0ae76261f772a1

PR #5292, opened 2021-05-04
DOCSP-15027 Introduces Versioned API in 5.0

jmd-mongo:


source/reference.txt
```diff
@@ -75,10 +75,13 @@ Reference
...
    Describes server sessions.
 
+:doc:`/reference/versioned-api`
+   Describes the Versioned API.
 
-.. seealso::
+.. seealso:: 
 
-   The
+   The :ref:`genindex` may provide useful insight into the
```
jeff-allen-mongo, line 83: I don't see this ref in this PR and I can't find it anywhere in the v5.0 branch. Should this be here?

jeff-allen-mongo, line 83: I don't see this `genindex` ref in this PR and I can't find it anywhere in the v5.0 branch. Should this be here?

PR #5294, opened 2021-05-05
DOCSP-15027 Introduces Versioned API in 5.0

jmd-mongo:


source/reference.txt
```diff
@@ -75,7 +75,9 @@ Reference
 :doc:`/reference/server-sessions`
    Describes server sessions.
 
-
+:doc:`/reference/versioned-api`
+   Describes server sessions.
```
jeff-allen-mongo, line 79: Describes server sessions? Is this a bad copy-paste?

jeff-allen-mongo (changes requested):

PR #5560, opened 2021-07-15
fixing wording

asya999:
shard keys don't have values 1, 2, 3...   and you define ranges using shard key, not zones, and to use location it has to be first field (not any one field) in the shard key


andf-mongodb:
Thanks Asya!

PR #5563, opened 2021-07-15
DOCSP-17572 mongosh options pass 3

andf-mongodb:


jeff-allen-mongo:
@andf-mongodb , this introduced some new build errors in the log. The problem is that the mongosh docs do not contain the `--ssl` options, since we are encouraging users to use the new `--tls` options instead.

For changes elsewhere in the docs where we mention the `-ssl` options, I just changed them to monospace.

andf-mongodb:
Thanks!

PR #5587, opened 2021-07-20
DOCSP-13855-mod-has-inconsistent-rounding

ianf-mongodb:


source/reference/operator/query/mod.txt
```diff
@@ -117,3 +117,31 @@ operator with an array that contains four elements:
    	"$err" : "bad query: BadValue malformed mod, too many elements",
    	"code" : 16810
    }
+
+Modulo with Decimal Divisors
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The ``modulo`` expression rounds down decimal input towards zero.
+
+The following examples demonstrates this behavior :
```
jeff-allen-mongo, line 126: [typo]

This should be "The following examples demonstrate..."

We also have an extra space between behavior and the colon.

source/reference/operator/query/mod.txt
```diff
@@ -117,3 +117,31 @@ operator with an array that contains four elements:
...
+   
+.. code-block:: javascript
```
jeff-allen-mongo, line 128: These examples are good, but this is not the most helpful format for users. Since this entire code block is copyable, users will expect to be able to copy and paste this block into their shell and have it run successfully. However, by including the input and output in the same code block, we run into issues.

One approach we could consider:

Place the input code block, followed by output code block, and label each code block accordingly. For example:
   
```
Input query:

.. code-block:: javascript

   db.inventory.find( { qty: { $mod: [ 4.0, 0 ] } } )

Results:

   .. code-block:: javascript
      :copyable: false
      
      { _id: 1, item: 'abc123', qty: 0 }
      { _id: 3, item: 'ijk123', qty: 12 }
```

And repeat this structure for each example.

source/reference/operator/query/mod.txt
```diff
@@ -117,3 +117,31 @@ operator with an array that contains four elements:
...
+
+   { _id: 1, item: 'abc123', qty: 0 }
+   { _id: 3, item: 'ijk123', qty: 12 }
+
+   db.inventory.find( { qty: { $mod: [ 4.99, 0 ] } } )
+   
+   { _id: 1, item: 'abc123', qty: 0 }
+   { _id: 3, item: 'ijk123', qty: 12 }
+
+Each query applies 4 to the mod function regardless of decimal points
```
jeff-allen-mongo, line 145: Since "4" is referring to a value in the code, we should monospace it:

```
``4``
```

source/reference/operator/query/mod.txt
```diff
@@ -117,3 +117,31 @@ operator with an array that contains four elements:
    	"$err" : "bad query: BadValue malformed mod, too many elements",
    	"code" : 16810
    }
+
+Modulo with Decimal Divisors
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The ``modulo`` expression rounds down decimal input towards zero.
```
jeff-allen-mongo, line 124: We should to this expression as `$mod` (not Modulo). As written, it's reading like we're introducing a new concept because up until now, we've been referring to this expression exclusively as `$mod`.

source/reference/operator/query/mod.txt
```diff
@@ -117,3 +117,31 @@ operator with an array that contains four elements:
...
+   { _id: 1, item: 'abc123', qty: 0 }
+   { _id: 3, item: 'ijk123', qty: 12 }
+
+   db.inventory.find( { qty: { $mod: [ 4.99, 0 ] } } )
+   
+   { _id: 1, item: 'abc123', qty: 0 }
+   { _id: 3, item: 'ijk123', qty: 12 }
+
+Each query applies 4 to the mod function regardless of decimal points 
+resulting in the same result set.
```
jeff-allen-mongo, line 146: This  sentence was a little unclear to me and reads like it might be missing a comma. Consider something like:

```
Each of the example queries apply ``4`` to the mod function regardless of decimal points, and result in the same output.
```

source/reference/operator/query/mod.txt
jeff-allen-mongo, line 124: Should we remove the word "down"? If I have a negative number, rounding toward zero would be rounding up, not rounding down.

jeff-allen-mongo (changes requested):

source/reference/operator/query/mod.txt
```diff
@@ -117,3 +117,31 @@ operator with an array that contains four elements:
    	"$err" : "bad query: BadValue malformed mod, too many elements",
    	"code" : 16810
    }
+
+Modulo with Decimal Divisors
```
jeff-allen-mongo, line 121: In the previous review, Nikita suggested changing this section title to "Floating Point Arguments". Did we want to incorporate that suggestion?

jeff-allen-mongo (approved):

PR #5614, opened 2021-07-25
Sort should be before not after group

asya999:


asya999:
This should be fixed in master and 5.0 - it was correct before then.

PR #5658, opened 2021-08-03
This page should exclude $search from this limitation.

ianf-mongodb:


source/includes/fact-agg-memory-limit.rst
```diff
@@ -8,6 +8,9 @@ space by using the :ref:`allowDiskUse <aggregate-cmd-allowDiskUse>`
 option to enable aggregation pipeline stages to write data to temporary
 files.
 
+The :manual:`search </text-search>` aggregation stage is not restricted
```
jeff-allen-mongo, line 11: Was there a specific request to link to this page instead of the [$search page](https://docs.mongodb.com/manual/reference/operator/aggregation/search/)?

I'd also suggest to say include the dollar sign in the link (`$search`, instead of `search`), if you use the following syntax, it will render in the monospace formatting, which may be preferable:
```
:pipeline:`$search`
```

PR #5680, opened 2021-08-06
Add blurb to 5.0 release notes for movePrimary error message

ianf-mongodb:


source/release-notes/5.0.txt
```diff
@@ -462,6 +462,13 @@ command to change the name of a sharded collection.
 
 .. _5.0-rel-notes-shell:
 
+``movePrimary`` Error Message For Writes During Operation
```
jeff-allen-mongo, line 465: Hey @ianf-mongodb  thanks for this update. We need this to be _before_  the `.. _5.0-rel-notes-shell:` reference on 463, because that reference points to the "Shell Changes" section.

ianf-mongodb: Hi Jeff, Thanks for pointing this out I have updated the PR.

PR #5709, opened 2021-08-18
DOCS-14257 db.collection.dropIndexes() 5.0 blurb  update

ianf-mongodb:


source/core/index-creation.txt
```diff
@@ -554,7 +554,9 @@ process:
   * - Mark the Index as Ready
 
     - The :binary:`~bin.mongod` updates the index metadata to 
-      mark the index as ready for use.
+      mark the index as ready for use.     
+      
+      .. _index-build-process-ready:
```
jeff-allen-mongo, line 559: I'm a little confused where this target is supposed to be. I clicked on the link on the `dropIndexes` page and felt like I was kind of thrown into the middle of a table and I wasn't sure where to look. Linking to a particular table row is tricky, especially given the currently wonky behavior of ref linking where you don't end up _exactly_ where you want and the top nav can obscure the target.

My suggestion would be to move this ref to the top of this section (Index Build Process, line 374), since then users will have a clear entry point for what we're trying to show them.

jeff-allen-mongo (approved):

source/core/index-creation.txt
jeff-allen-mongo: Comment resolved via Slack.

PR #5722, opened 2021-08-22
minor fixes to multikey-index-bounds.txt and index-wildcard.txt

RafaelGreen1:


source/core/index-wildcard.txt
```diff
@@ -439,15 +439,15 @@ Consider the following wildcard index on the ``employees`` collection:
 
 .. code-block:: javascript
 
-   db.products.createIndex( { "$**" : 1 } )
+   db.employees.createIndex( { "$**" : 1 } )
```
jeff-allen-mongo, line 442: Thank you for this fix!

source/core/multikey-index-bounds.txt
```diff
@@ -349,10 +349,10 @@ field ``ratings`` to require that the array contains at least one
 
 Taking the predicates separately:
 
-- the bounds for the ``score: { $lte: 5 }`` predicate is ``[ -Infinity, 5
-  ]``;
+- the bounds for the ``score: { $lte: 5 }`` predicate are ``[ [ -Infinity, 5
```
jeff-allen-mongo, line 352: I agree with changing "is" to "are", but what is the motivation for the double square brackets?

RafaelGreen1 (commented):
it appears like this in all other places on this page. search "[ [" in this page to see the other occurrences.
it makes sense to me because bounds are arrays of arrays. for example: [ [1,3], [8,10] ]

source/core/multikey-index-bounds.txt
RafaelGreen1: It apprears like this in all other places on this page. Search "[ [" to see the other occurances.
It makes sense to me because bounds are arrays of arrays. For example [ [1,3], [8,10] ]

PR #5723, opened 2021-08-23
DOCS-14756: Update ensure-indexes-fit-ram.txt

RafaelGreen1:
the calculation from Bytes to GB is wrong


jeff-allen-mongo:
Thank you @RafaelGreen1 !

PR #5767, opened 2021-08-30
(DOCSP-18425): Add Go to Driver Support for DBRefs table

jeff-allen-mongo:


matthewdale (approved):
Looks good 👍

source/reference/database-references.txt
```diff
@@ -227,6 +227,12 @@ Driver Support for DBRefs
 
      - You can traverse references manually.
 
+   * - Go
```
matthewdale, line 230: Should these be in alphabetical order? If so, this should be above the "Haskell" line.

matthewdale (commented):
Just gave this a 👍 , but I actually have one question about table ordering.

source/reference/database-references.txt
jeff-allen-mongo: Oh my mistake you're right! Will fix. Cheers.

PR #5780, opened 2021-09-01
DOCS-13309 short circuit evaluation

jason-price-mongodb:


source/reference/operator/aggregation/and.txt
```diff
@@ -66,10 +63,27 @@ evaluation after encountering the first ``false`` expression.
 
      - ``false``
 
+Error Handling
+--------------
```
jeff-allen-mongo, line 67: Hey Jason, sorry for not being clearer in my comment. I meant to make these sub-headings of the "Behavior" section above.

source/reference/operator/aggregation/or.txt
```diff
@@ -60,6 +57,23 @@ evaluation after encountering the first ``true`` expression.
 
      - ``false``
 
+Error Handling
+--------------
```
jeff-allen-mongo, line 61: Suggest converting to a sub-heading of the Behavior section above.

jeff-allen-mongo: (discussion resolved via Slack. no action needed).

PR #5783, opened 2021-09-01
DOCS-14433 config.transactions read concern disallowed

ianf-mongodb:


source/includes/fact-5.0-non-transactional-config-reads.rst
```diff
@@ -0,0 +1,10 @@
+Starting in MongoDB 5.0, non-transaction reads are not allowed on
+:ref:`config.transactions<config.transactions>`
```
jeff-allen-mongo, line 2: Two things:

1) Please use
```
:data:`config.transactions`
```
for this link instead of a `ref` link.

2) I think we should say:
```
not allowed on the :data:`config.transactions` collection with the following...
```

source/includes/fact-5.0-non-transactional-config-reads.rst
```diff
@@ -0,0 +1,10 @@
+Starting in MongoDB 5.0, non-transaction reads are not allowed on
+:ref:`config.transactions<config.transactions>` 
+with the following read concerns and considerations:
+
+- :ref:`read concern "snapshot"<readconcern-snapshot>`
+- :ref:`read concern "majority"<readconcern-majority>` and the 
+  :ref:`afterClusterTime<afterClusterTime>` option is set
+- When using a `MongoDB Driver <https://docs.mongodb.com/drivers/?tck=docs_server>`_  
+  and :ref:`read concern "majority"<readconcern-majority>` 
+  within a :ref:`causally consistent session<sessions>`
```
jeff-allen-mongo, line 10: Please update these to use the `:readconcern:` links instead of `refs`

source/reference/read-concern-majority.txt
```diff
@@ -10,6 +10,8 @@ Read Concern ``"majority"``
    :depth: 1
    :class: singlecol
 
+.. _readconcern-majority:
```
jeff-allen-mongo, line 13: You don't need to add these refs. `.. readconcern:: "majority"` you can already point to this place in the docs by using:

```
:readconcern:`"majority"`
```

jeff-allen-mongo (changes requested):

source/includes/fact-5.0-non-transactional-config-reads.rst
```diff
@@ -0,0 +1,10 @@
+Starting in MongoDB 5.0, non-transaction reads are not allowed on
+:ref:`config.transactions<config.transactions>` 
+with the following read concerns and considerations:
+
+- :ref:`read concern "snapshot"<readconcern-snapshot>`
+- :ref:`read concern "majority"<readconcern-majority>` and the 
+  :ref:`afterClusterTime<afterClusterTime>` option is set
+- When using a `MongoDB Driver <https://docs.mongodb.com/drivers/?tck=docs_server>`_
```
jeff-allen-mongo, line 8: You can do:

```
:driver:`MongoDB Driver </>`
```

This is defined in the snooty rst spec: https://github.com/mongodb/snooty-parser/blob/master/snooty/rstspec.toml

source/includes/fact-5.0-non-transactional-config-reads.rst
```diff
@@ -0,0 +1,10 @@
+Starting in MongoDB 5.0, non-transaction reads are not allowed on
+:ref:`config.transactions<config.transactions>` 
+with the following read concerns and considerations:
```
jeff-allen-mongo, line 3: [minor] "options" may be better than "considerations" here

source/reference/config-database.txt
```diff
@@ -524,6 +526,8 @@ instance.
      the ``system.sessions`` collection into at least 1024 chunks and
      distributes the chunks uniformly across shards in the cluster.
 
+.. _config.transactions:
```
jeff-allen-mongo, line 529: I think we can remove this as well.

jeff-allen-mongo (approved):

PR #5787, opened 2021-09-02
DOCS-14695-Upgrading language strength with regards to hidden delayed…

ianf-mongodb:


source/includes/important-delayed-replica-set-members.rst
```diff
@@ -0,0 +1,27 @@
...
+
+   If your replica set contains :doc:`delayed members
+   </core/replica-set-delayed-member>` ensure that the delayed
+   members are hidden and non-voting. 
+
+   Hiding delayed replica set members prevents applications from seeing 
+   and querying delayed data without a direct connection to that member.
+   Making delayed replica set members non-voting means they will not 
+   count towards acknowledging write operations with read concern 
+   :readconcern:`"majority"`.
```
jeff-allen-mongo, line 11: Should this be "write concern majority" ? Instead of read concern majority?

source/includes/important-delayed-replica-set-members.rst
```diff
@@ -0,0 +1,27 @@
...
+
+   If you do not hide delayed members and one or more nodes 
+   becomes unavailable, the replica set has to wait for the delayed 
+   member and the commit point lags. A lagged commit point can lead to
+   performance issues.
```
jeff-allen-mongo, line 16: There is some shorthand and tone in this section and following example that doesn't really match the tone of the docs. We don't really use "config" as shorthand for "configuration", or "mins" as short for "minutes". 


Consider something like:

```
For example, consider a Primary-Secondary-Delayed replica set
configuration where the delayed secondary is voting with a 10
minute delay.

With one non-delayed secondary unavailable, the degraded configuration
of Primary-Delayed must wait at least 10 minutes to acknowledge a write
operation with :writeconcern:`"majority"`.

The majority commit point will take longer to advance, leading to cache
pressure similar performance issues with...
```

source/includes/important-delayed-replica-set-members.rst
```diff
@@ -0,0 +1,27 @@
...
+
+   For example:
+
+   - Primary-Secondary-Delayed replica set config where delayed 
+     secondary is voting with 10 mins delay.
```
jeff-allen-mongo, line 21: This bullet is a sentence fragment which reads a little awkwardly. See suggestion in the comment above.

source/includes/important-delayed-replica-set-members.rst
```diff
@@ -0,0 +1,27 @@
...
+   - With one non-delayed secondary unavailable, the degraded config of
+     Primary-Delayed will have to wait at least 10 minutes for a write 
+     concern with majority to be acknowledged.
+   - The majority commit point will take longer to advance leading to 
+     cache pressure similar to performance issues with a PSA replica 
+     set.
```
jeff-allen-mongo, line 27: Can we end this thought with something like:

> For more information on the majority commit point, see [Causal Consistency and Read and Write Concerns](https://docs.mongodb.com/manual/core/causal-consistency-read-write-concerns/)

I think "majority commit point" is a little confusing and providing that link for context may be useful.

I'm also not sure I understand the bit about a PSA replica set. We haven't mentioned them at all in this page and I feel like I'm missing some context. Is there anywhere we can link to in the docs that explains more about these performance issues? It kind of seems to come out of nowhere. We should also expand the acronym since users might not know what we mean by a PSA replica set.

jeff-allen-mongo:
@ianf-mongodb , just pinging in case these changes fell off your radar.

source/includes/important-delayed-replica-set-members.rst
ianf-mongodb: Updated to write concern.

source/includes/important-delayed-replica-set-members.rst
ianf-mongodb: Updated.

source/includes/important-delayed-replica-set-members.rst
ianf-mongodb: Updated.

source/includes/important-delayed-replica-set-members.rst
ianf-mongodb: Updated / added a link to the replica set maintenance tutorial, which goes into more context of resolving the performance issues detailed with the commit point.

source/includes/important-delayed-replica-set-members.rst
```diff
@@ -0,0 +1,35 @@
...
+   members are hidden and non-voting. 
+
+   Hiding delayed replica set members prevents applications from seeing 
+   and querying delayed data without a direct connection to that member.
+   Making delayed replica set members non-voting means they will not 
+   count towards acknowledging write operations with write concern 
+   :writeconcern:`"majority"`.
+
+   If you do not hide delayed members and one or more nodes 
+   becomes unavailable, the replica set has to wait for the delayed
```
jeff-allen-mongo, line 14:
```suggestion
   become unavailable, the replica set has to wait for the delayed 
```

source/core/replica-set-delayed-member.txt
```diff
@@ -39,12 +39,14 @@ Delayed members:
   members. Set the priority to 0 to prevent a delayed member from
   becoming primary.
 
-- **Should be** :ref:`hidden <replica-set-hidden-members>`
+- **Must be** :ref:`hidden <replica-set-hidden-members>`
   members. Always prevent applications from seeing and querying
   delayed members.
 
 - *Do* vote in :term:`elections <election>` for primary, if :rsconf:`members[n].votes` is set to 1.
```
jeff-allen-mongo, line 46: I'm a little confused because here we are saying that the delayed members vote in elections for the primary, but then immediately below we say that users should ensure that they are non-voting. It seems slightly contradictory.

Do you think we should we should add a sentence to this bullet that says something like:

```
Ensuring that delayed members are non-voting by
setting :rsconf:`members[n].votes` to 0 can help improve performance.
```

source/includes/important-delayed-replica-set-members.rst
```diff
@@ -0,0 +1,35 @@
...
+   operation with :writeconcern:`"majority"`.
+
+   The majority commit point will take longer to advance, leading to cache
+   pressure similar performance issues with a
+   :ref:`Primary with a Secondary and an Arbiter<rs-architecture-psa>`
+   (PSA) replica set.
+
+   For more information on the majority commit point, see 
+   :doc:`Causal Consistency and Read and Write Concerns
+   </core/causal-consistency-read-write-concerns>` for additional details
```
jeff-allen-mongo, line 33: This should be two separate sentences?

jeff-allen-mongo (changes requested):
A few more thoughts here.

source/includes/important-delayed-replica-set-members.rst
ianf-mongodb: Separated in latest.

source/core/replica-set-delayed-member.txt
ianf-mongodb: Im reading this as a warning to the user, in other words the first two bullets, which start with **Must be** are the recommendation. The *Do* is a warning if you do not address the rsconfig`members[n].votes` setting. I think the extra blurb does add value.

source/includes/important-delayed-replica-set-members.rst
```diff
@@ -0,0 +1,36 @@
...
+   minute delay.
+
+   With one non-delayed secondary unavailable, the degraded configuration
+   of Primary-Delayed must wait at least 10 minutes to acknowledge a write
+   operation with :writeconcern:`"majority"`.
+
+   The majority commit point will take longer to advance, leading to cache
+   pressure similar performance issues with a
+   :ref:`Primary with a Secondary and an Arbiter<rs-architecture-psa>`
+   (PSA) replica set.
```
jeff-allen-mongo, line 29: Can we combine these into a single paragraph? This block is very big because we have so many paragraphs so I'm looking for ways to combine some of them.

source/includes/important-delayed-replica-set-members.rst
```diff
@@ -0,0 +1,36 @@
...
+
+   For more information on the majority commit point, see 
+   :doc:`Causal Consistency and Read and Write Concerns
+   </core/causal-consistency-read-write-concerns>`. 
+   
+   For additional details on resolving performance issues see the
+   :ref:`replica set maintenance tutorial<performance-issues-psa>`.
```
jeff-allen-mongo, line 36: Similarly, we could combine these into a single paragraph as well.

jeff-allen-mongo (commented):
Thanks for making the edits. Just two more small things.

jeff-allen-mongo (approved):

PR #5842, opened 2021-09-16
DOCS-14281 add release notes and update partial index restrictions

ianf-mongodb:


source/reference/method/db.collection.createIndexes.txt
```diff
@@ -270,7 +271,8 @@ otherwise specified:
 
      - document
 
-     - Optional. If specified, the indexes only reference documents that match the
+     - .. _partialFilterExpression:
+       Optional. If specified, the indexes only reference documents that match the
```
jeff-allen-mongo, line 275: There should be a blank line between where we declare the ref and the content.

```suggestion
     - .. _partialFilterExpression:
   
       Optional. If specified, the indexes only reference documents that match the
```

source/reference/method/db.collection.createIndexes.txt
```diff
@@ -38,7 +38,8 @@ Definition
    
         - document
    
-        - An array containing index specification documents. Each document
+        - .. _key_patterns:
+          An array containing index specification documents. Each document
```
jeff-allen-mongo, line 42: There should be a blank line between where we declare the ref and the content.

```suggestion
        - .. _key_patterns:
        
          An array containing index specification documents. Each document
```

PR #5873, opened 2021-09-24
DOCS-13871 Add REPL output to serverStatus command

ianf-mongodb:


source/release-notes/5.0.txt
```diff
@@ -895,6 +895,13 @@ Security Counter
     which reports the number of times a :dbcommand:`hello` request
     includes a valid :data:`hello.saslSupportedMechs` field.
 
+Repl
+  - :serverstatus:`repl` now includes a
+    :ref:`primaryOnlyServices<server-status-repl>` document which
```
jeff-allen-mongo, line 900: Two things:

- Should this go into the "Replication Metrics" section above this one, rather than creating a new "Repl" section?
- Should we link directly to the "primaryOnlyServices" document? As in:
```
:serverstatus:`~repl.primaryOnlyServices`
```

The tilde at the beginning trims everything before the field name, so it should render as "primaryOnlyServices" rather than "repl.primaryOnlyServices".

jeff-allen-mongo: This should be "document 'that' contains", not 'which'.

jeff-allen-mongo (commented):
Just a couple questions before merge.

source/release-notes/5.0.txt
ianf-mongodb: I updated the link and the "which" -> "that", as for the repl output I feel it is more appropriate to leave it as is I feel like this would have come out in TR if it were not the case. I attached [a copy of serverStatus]([https://jira.mongodb.org/browse/DOCS-13871]) output to the ticket where metrics shows under a different document hierarchy.

PR #5923, opened 2021-10-05
DOCSP-14047 add note about mongoshell to page

ianf-mongodb:


source/faq/fundamentals.txt
```diff
@@ -44,35 +44,51 @@ the columns in a relational database table.
 How do I create a database and a collection?
 --------------------------------------------
 
+.. note:: 
+
+   You can enter the commands referenced in this FAQ by using 
+   the :binary:`MongoDB Shell <~bin.mongo>`. The MongoDB Shell is an
```
jeff-allen-mongo, line 50: We're linking to the legacy mongo shell but the text refers to the current MongoDB Shell.

ianf-mongodb: updated.

PR #5924, opened 2021-10-05
DOCSP-14047 add note about mongoshell to page

ianf-mongodb:


source/faq/fundamentals.txt
```diff
@@ -41,35 +41,51 @@ the columns in a relational database table.
 How do I create a database and a collection?
 --------------------------------------------
 
+.. note:: 
+
+   You can enter the commands referenced in this FAQ by using 
+   the :doc:`Mongo Shell </mongo/>`. The MongoDB Shell is an
```
jeff-allen-mongo, line 47: If we're referring to the legacy shell, "mongo" should be lower case.

source/faq/fundamentals.txt
```diff
@@ -41,35 +41,51 @@ the columns in a relational database table.
 How do I create a database and a collection?
 --------------------------------------------
 
+.. note:: 
+
+   You can enter the commands referenced in this FAQ by using 
+   the :doc:`Mongo Shell </mongo/>`. The MongoDB Shell is an 
+   interactive JavaScript interface to MongoDB. You can use the MongoDB
```
jeff-allen-mongo, line 48: We need to be consistent in the terms here. We used "mongo Shell" above but are using MongoDB Shell here.

jeff-allen-mongo (changes requested):

source/faq/fundamentals.txt
jeff-allen-mongo: We may want to use

```
:binary:`~bin.mongo`
```

to link to the legacy shell.

ianf-mongodb: Thats what I was refering to in my comments earlier, that link was not rendering a valid link in 4.2 or 4.0. I did think about adding a :4.2: to conf.py (currently there is not one). But for some reason the binary was not working in that branch.

PR #5964, opened 2021-10-08
DOCSP-18495 link to sharding in agg ebook

ghost:


source/sharding.txt
```diff
@@ -390,6 +390,19 @@ available on sharded clusters.
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+For more information on how sharding works with
+:ref:`aggregations <aggregation-framework>`, read the sharding chapter
+in the `Practical MongoDB Aggregations
+<https://www.practical-mongodb-aggregations.com/guides/sharding.html>`__
+e-book.
+
+
 For details, see:
```
jeff-allen-mongo, line 406: The way this is staged, we have the "For details, see: " as part of the new "E-book" section. As a reader, since these are part of the E-book section, i now expect these to be links within the e-book.

We should consider adding a sub-heading for these docs links like "Additional Information" or similar for these links.

PR #5977, opened 2021-10-12
DOCS-14840 5.0.4 setWindowFields restriction updates

ianf-mongodb:


source/reference/operator/aggregation/setWindowFields.txt
```diff
@@ -242,6 +242,12 @@ Restrictions
 
 Restrictions for the :pipeline:`$setWindowFields` stage:
 
+- Starting in MongoDB 5.1 (and 5.0.4), the :pipeline:`$setWindowFields`
```
jeff-allen-mongo, line 245: Hey @ianf-mongodb , when backporting, we should only mention the versions that apply to the version the reader is on and previous.

```suggestion
- Starting in MongoDB 5.0.4, the :pipeline:`$setWindowFields` 
```

jeff-allen-mongo (changes requested):

PR #5997, opened 2021-10-14
DOCS-14593 update remainingOperationTimeEstimate blurb

ianf-mongodb:


source/release-notes/5.0.txt
```diff
@@ -524,6 +524,14 @@ uneven data distribution. Starting in MongoDB 5.0, you can use the
 <change-a-shard-key>` for a collection to change the distribution of
 your data across your cluster.
 
+``currentOp`` Output Change
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Starting in MongoDB 5.0, the 
+:data:`$currentOp.remainingOperationTimeEstimated` metric is only 
+present on the :ref:`recipient shard <resharding-process-details>`
+when a :ref:`resharding operation <sharding-resharding>` is taking place.
```
jeff-allen-mongo, line 533: I think this should go on the compatibility changes, rather than the the main release notes.

If users are relying on this metric being available on all shards rather than only recipient shards, then their applications may behave unexpectedly or break. Users would ideally find that kind of information under "compatibility changes" to understand all potentially breaking changes that come with upgrading to 5.0.

PR #6067, opened 2021-11-01
DOCS-14821 Adding metadata flag context to validate()

ianf-mongodb:


source/includes/fact-validate-repair-option.rst
```diff
@@ -29,7 +29,8 @@ The repair fixes these issues:
 - If corrupt documents with invalid BSON data are found, the documents
   are removed.
 
-.. seealso:: :option:`--repair <mongod --repair>` option for
-   :binary:`~bin.mongod`
+.. seealso::
```
ianf-mongodb, line 32: FYI : This was a fix for a see also directive that was showing empty on the page. Cleaned up with this effort.

source/includes/fact-validate-metadata.rst
```diff
@@ -0,0 +1,33 @@
...
+find invalid indexes more quickly.
+
+If there is an invalid index detected, the validate command will prompt
+you to use the :dbcommand:`collMod` command to remove invalid indexes.
+
+.. code-block:: javascript
+
+   db.runCommand( { collMod: <collectionName> } )
+
+.. versionadded:: 5.0
```
jeff-allen-mongo, line 33: Hey @ianf-mongodb , this looks like it's only available in 5.0.4. I think we need to be explicit about that, otherwise users running 5.0.3 might expect this feature.

source/release-notes/5.0.txt
```diff
@@ -761,6 +761,14 @@ The :dbcommand:`validate` command and :method:`db.collection.validate()`
 helper method also return a new :data:`~validate.repaired` boolean value
 that is ``true`` if the collection was repaired.
 
+Metadata Option Added for ``validate`` Command
```
jeff-allen-mongo, line 764: Since this isn't applicable to all minor versions of 5.0 (it's 5.0.4+), we should put this in the 5.1 release notes and say:

"Starting in MongoDB 5.1 (and MongoDB 5.0.4), ..."

jeff-allen-mongo (changes requested):

source/release-notes/5.0.txt
ianf-mongodb: Updated and moved to 5.1 notes

source/includes/fact-validate-metadata.rst
ianf-mongodb: Updated to ".. versionadded:: 5.0.4"

jeff-allen-mongo (approved):

PR #6073, opened 2021-11-03
DOCS-14779 Add context around 5.1 let usage in geoNear

ianf-mongodb:


source/reference/operator/aggregation/geoNear.txt
```diff
@@ -204,8 +204,147 @@ When using :pipeline:`$geoNear`, consider that:
...
+- Starting in MongoDB 5.1, the ``near`` parameter supports the 
+  :ref:`let option <geoNear_let_example>` and
+  :ref:`bound let option <geoNear_bounded_let_example>`.
+
+Examples
+--------
+
+.. _geoNear_let_example:
+
+$geoNear with the ``let`` option
```
jeff-allen-mongo, line 216: I think the maximum and minimum distance examples should still be the first examples in this section. I'm worried that if we put the `let` option first, we're introducing a fairly abstract / complex example as the first one on the page. Let's make sure that the simplest, most straight-forward examples appear first when possible.

source/reference/operator/aggregation/geoNear.txt
```diff
@@ -204,8 +204,147 @@ When using :pipeline:`$geoNear`, consider that:
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+In this example:
+
+- The ``let`` option is used to set an array value of 
+  ``[-73.99279,40.719296]`` to the variable ``$pt``.
```
jeff-allen-mongo, line 222: What is the significance of the name "pt"? Is it short for "point", or is it an acronym of some kind?

source/reference/operator/aggregation/geoNear.txt
```diff
@@ -204,8 +204,147 @@ When using :pipeline:`$geoNear`, consider that:
...
+
+- ``$pt`` is a let option to the ``near`` parameter to the ``$geoNear`` 
+  stage.
```
jeff-allen-mongo, line 225: This phrasing here is a little off.

```suggestion
- ``$pt`` is specified as a let option to the ``near`` parameter in the ``$geoNear`` 
  stage.
```

source/reference/operator/aggregation/geoNear.txt
```diff
@@ -204,8 +204,147 @@ When using :pipeline:`$geoNear`, consider that:
...
+      name: 'Central Park',
+      location: { type: 'Point', coordinates: [ -73.97, 40.77 ] },
+      category: 'Parks',
+      distance: 0.0009348548688841822,
+      dist: { location: { type: 'Point', coordinates: [ -73.97, 40.77 ] } }
+   }
+
+.. _geoNear_bounded_let_example:
+
+$geoNear With Bound ``let`` Option
```
jeff-allen-mongo, line 273:
```suggestion
$geoNear with Bound ``let`` Option
```

source/reference/operator/aggregation/geoNear.txt
```diff
@@ -204,8 +204,147 @@ When using :pipeline:`$geoNear`, consider that:
...
+                        distanceField: "distance"
+                     }
+               }
+               ],
+            as: "joinedField"
+         }
+      },
+      { $match: { name: "Sara D. Roosevelt Park" } }
+   ]
+   );
```
jeff-allen-mongo, line 309: The indentation here is a little off. We (generally) need the closing brackets to align with their opening counterparts. Suggest something like:

```suggestion
   db.places.aggregate( [
   {
      $lookup: {
         from: "places",
         let: { pt: "$location" },
         pipeline: [
            {
               $geoNear: {
                  near: "$$pt",
                  distanceField: "distance"
               }
            }
         ],
         as: "joinedField"
      }
   },
   { 
      $match: { name: "Sara D. Roosevelt Park" }
   }
] );
```

jeff-allen-mongo (changes requested):
Hey @ianf-mongodb , just a few small things to address prior to merge. Thanks!

source/reference/operator/aggregation/geoNear.txt
```diff
@@ -204,8 +204,147 @@ When using :pipeline:`$geoNear`, consider that:
...
+            "spherical":true
+         }
+      }
+   ],
+   {
+      "let":{ "pt": [ -73.99279, 40.719296 ] }
+   }
+   )
+
+The aggregation returns the following:
```
jeff-allen-mongo, line 249: Can we explain a little bit what the results mean? The other examples on this page have explanations like:

> The following aggregation uses $geoNear to find documents with a location at most 2 meters from the center [ -73.99279 , 40.719296 ] and category equal to Parks.

Users might have an easier time following these examples if we briefly explain what the goal of the pipeline we've created is.

source/reference/operator/aggregation/geoNear.txt
ianf-mongodb: Updated.

source/reference/operator/aggregation/geoNear.txt
ianf-mongodb: Its just the variable name for the point array.

source/reference/operator/aggregation/geoNear.txt
ianf-mongodb: Updated.

source/reference/operator/aggregation/geoNear.txt
ianf-mongodb: Updated.

source/reference/operator/aggregation/geoNear.txt
ianf-mongodb: Should the { (previously line #290) for $lookup also come down to the next line to match the indent? See updated.

source/reference/operator/aggregation/geoNear.txt
ianf-mongodb: I specified this above each of the queries : 

In this example, :pipeline:`$lookup` uses:

- ``let`` to define ``$pt``.
- :pipeline:`$geoNear` in the ``pipeline``.
- ``$pt`` to define ``near`` in the :pipeline:`$geoNear` pipeline stage.

In this example:

- The ``let`` option is used to set an array value of 
  ``[-73.99279,40.719296]`` to the variable ``$pt``.

- ``$pt`` is specified as a let option to the ``near`` parameter in the 
  ``$geoNear`` stage.
  
  Do you mean explictly changing "the aggregation returns the following" to describe the pipeline? I tried to  copy the existing format as much as possible.

ianf-mongodb (commented):
Updated + responded.

source/reference/operator/aggregation/geoNear.txt
jeff-allen-mongo: Thanks for clarifying. fwiw I don't love this as a variable name because it wasn't immediately clear to me what the significance of it was, but we can leave it alone.

source/reference/operator/aggregation/geoNear.txt
```diff
@@ -287,6 +294,136 @@ equal to ``Parks``.
...
+      }
+   ] );
+
+The aggregation returns the following:
+
+.. code-block:: javascript
+   :copyable: false
+
+   {
+    _id: ObjectId("61715cf9b0c1d171bb498fd7"),
```
jeff-allen-mongo, line 398: Do these objects need to be indented 1-2 more spaces? I'm wondering why the syntax highlighting isn't working in this code-block.

source/reference/operator/aggregation/geoNear.txt
jeff-allen-mongo: Yeah I saw, but my point was more that we've explained what the parameters are but not how they affect the end result. I think if we adjusted:

> The aggregation returns the following:

To be something like:

```
The aggregation returns all documents with: 

- A location at most 2 meters from the point defined in  the ``let`` variable
- A ``category`` equal to ``Parks``.

.. code-block::
```

it could help paint a clearer picture. The examples which were on the page prior to this change do have an explanation like this. However, if you think the current approach is clear enough we can leave this alone.

jeff-allen-mongo (commented):
Thanks for the updates and comments @ianf-mongodb. Responded.

source/reference/operator/aggregation/geoNear.txt
ianf-mongodb: Yes in addition to the syntax not picking up the copy button still renders when copyable : false is used. Corrected.

ianf-mongodb:
Hey Jeff added the context to the example queries and correct the code box so it is displaying the syntax highlighting correctly now. @jeff-allen-mongo

jeff-allen-mongo:
Thanks @ianf-mongodb , did you see this [comment](https://github.com/mongodb/docs/pull/6073/files#r742857546) from my last review? I don't see the update for that one.

ianf-mongodb (commented):
Committed suggestion.

jeff-allen-mongo (approved):

PR #6145, opened 2021-11-15
DOCSP-18626 release note for snippets.

davemungo:


addaleax (approved):

jeff-allen-mongo:
Hey @davemungo , sorry to interject, I see this has already been merged but just curious why we're also including this feature on the core server release notes. We have dedicated release notes for the shell, so I'm unsure why we're also highlighting this feature here.

This feature wasn't released with MDB 5.0.0 specifically. This page is really meant to be for new features / changes in core MongoDB 5.0.0.

Is the goal to provide more visibility for snippets? Why not just only include this on the mongosh release notes?

PR #6181, opened 2021-11-22
fix consideration note for 2d indexes

Sorosliu1029:
image

- fix: `2dsphere` index
- fix: duplicated 'index'


jeff-allen-mongo:
Thank you @Sorosliu1029 !

PR #6183, opened 2021-12-13
DOCS-14980 updated lookup documentation

LakshmiLavanyaKasturi:
Removed the unsharded qualifier for lookup collections. 
$graphlookup text would need to be verified.


jeff-allen-mongo:
Hi @LakshmiLavanyaKasturi, thanks very much for your contribution! One of our writers included this change along with a few others and we are going through review internally. You can expect to see these changes live shortly.

PR #6190, opened 2022-02-03
Update Makefile

GuruPKK:


Makefile
```diff
@@ -5,6 +5,16 @@ PRODUCTION_URL="https://docs.mongodb.com"
 STAGING_BUCKET=docs-mongodb-org-stg
 PRODUCTION_BUCKET=docs-mongodb-org-prd
 
+ifeq ($(ENV), 'dotcom_stg')
```
jdestefano-mongo, line 8: Do you think we should avoid calling this "ENV" to avoid conflict with the unix `env` command? Maybe something like docsenv or publishenv?

Makefile
```diff
@@ -5,6 +5,16 @@ PRODUCTION_URL="https://docs.mongodb.com"
...
+	STAGING_URL="https://docs-mongodborg-staging.corp.mongodb.com"
+	STAGING_BUCKET=docs-mongodb-org-dotcomstg
+endif
+
+ifeq ($(ENV), 'dotcom_prd')
```
jdestefano-mongo, line 13: Writers will want to stage and publish to the same environment. Let's have one variable check for "dotcom" and set both the staging and prod URLs/buckets in one shot.

Makefile
```diff
@@ -5,6 +5,16 @@ PRODUCTION_URL="https://docs.mongodb.com"
 STAGING_BUCKET=docs-mongodb-org-stg
 PRODUCTION_BUCKET=docs-mongodb-org-prd
 
+ifeq ($(ENV), 'dotcom_stg')
+	STAGING_URL="https://docs-mongodborg-staging.corp.mongodb.com"
```
jdestefano-mongo, line 9: Per discussion staging URL needs to change.

jdestefano-mongo (changes requested):
@GuruPKK Thanks for the PR, requesting some changes. Cheers

PR #6212, opened 2022-06-28
DOP-3026: publish manifests to correct bucket/path

schmalliso:


schmalliso:
@jeff-allen-mongo  Can you review? Just fixing the Makefile for v4.2 so that deploys of it put search in the right bucket.

PR #6311, opened 2026-09-22
Fix typo in content/manual/manual/source/reference/parameters.txt

Rawal27:
## Summary

* Correct `affect` to `effect` in the `httpVerboseLogging` parameter documentation in `parameters.txt`.
* `effect` is the correct noun in this context, referring to an impact or result.
* No technical or behavioral changes are introduced.


AKzar1el (approved):
`effect` is the correct noun here, and the change is limited to that one sentence in the `httpVerboseLogging` parameter docs. I checked the surrounding entry and there’s no behavior or formatting change involved. Looks good.

